PR #13, opened 2021-01-20
feat: :sparkles: adds styled-components

AdrianValdes:
Adds styled components from `npm install --save styled-components` and also adds support for Typescript with `npm install @types/styled-components`. To do: Maybe install a babel plugin (`npm install --save-dev babel-plugin-styled-components`) or a a Typescript plugin (`npm i typescript-plugin-styled-components --dev`). I didn't do it because there are issues opened and apparently all those pluggins don't work together.


AdrianValdes:
Hi Christos,

I suppose there is work more work to do in this issue. But I am not sure how to test if the made changes work correctly. I did the same steps in another project and I was able to create styled-components to render them in the UI with React, but I don't know how to create new stories.

I have read a lot of documentation, but the majority of the articles are not updated all to things differently.

paschalidi:
Perhaps in this pull request we could refactor the components and have them moved in their own folders. 
Then you would have the `packages/ui/stories/Button`, `packages/ui/stories/Header`. In each component folder you will have three files. a) the story, b) the component, c) the `componentName.styles.js` which will contain the file.

What you could do to test that the styled components work, which most likely they do already, is to replace the button.css with a `button.styles.js` where you replace the css with the styled components stuff.

Wanna give it a shot? And if anything isnt clear you let me know alright?

AdrianValdes:
Yes, yes, I want to give it a try! I will let you know.

AdrianValdes:
Hi, I have been busy with the class today. But I will have a time to keep working on this later. Just one question to be sure, when you say:

>  Then you would have the packages/ui/stories/Button, packages/ui/stories/Header etc. 

Are you placing the `stories` outside the `src` on purpose? I checked another project of yours and you have the `stories` inside the `src` folder. So at the end would be `packages/ui/src/stories/Button`, but maybe that is not what you want for this project.

Let me know, and I am sorry if I am making you work twice. I look forward to be more of an asset soon : )

paschalidi:
>Are you placing the stories outside the src on purpose? I checked another project of yours and you have the stories inside the src folder. So at the end would be packages/ui/src/stories/Button, but maybe that is not what you want for this project. 

Nope I was mistaken sorry.

AdrianValdes:
Hi, I refactored the code to use styled-components. Like you probably realized already, I copied the code for a Button you already created. I only did this because I wanted to check everything is working and I am not sure how to use some features like `storiesOf`, not out of laziness. 

I didn't refactor the Page and Header stories for two reasons: 
1. Again I wanted to check that everything works with the Button story first.
2. I am not sure we would need these two components. So maybe we delete them?

To do:
1. Maybe delete the `Introduction.stories.mdx` file.
2. Place the Colors object in its own folder and then export it, (and use or own colors palette maybe)
3. Check the properties with type `any` on the `ButtonProps` interface.

paschalidi:
Will be looking into this tomorrow I believe. I would say what you can do in the meantime is to figure out those `any` on the button. I believe we would want to avoid annotating with `any` whenever this is possible. Do some reading about it and find the best way to represent `children` and the onClick `function`

paschalidi:
Also you right about Colors. Thats a good idea, please go on and export it on its own.

paschalidi:
Lastly yes you can delete anything else apart from the button. We can only keep the button for now.

packages/ui/src/stories/Button/Button.tsx
```diff
@@ -0,0 +1,50 @@
+import React from "react";
+/* import './button.css'; */
```
paschalidi, line 2: I think you might not need this

packages/ui/src/stories/Button/Button.stories.tsx
```diff
@@ -0,0 +1,20 @@
...
+  .add("primary", () => <Button>Call to Action</Button>)
+  .add("primary disabled", () => <Button disabled>Call to Action</Button>)
+  .add("secondary", () => <Button type="secondary">Call to Action</Button>)
+  .add("secondary disabled", () => (
+    <Button disabled type="secondary">
+      Call to Action
+    </Button>
+  ))
+  .add("onClick handler", () => (
+    // eslint-disable-next-line no-alert
```
paschalidi, line 16: This is okay here since this is simply a story, but keep in mind that in actual code we would want to avoid disabling eslint 99.9 percent of the time :)

packages/ui/.storybook/main.js
```diff
@@ -4,5 +4,6 @@ module.exports = {
     "@storybook/addon-links",
     "@storybook/addon-essentials",
     "@storybook/preset-create-react-app",
+    "@storybook/addon-controls",
```
paschalidi, line 7: What would that help us with?

packages/ui/src/stories/Button/Button.tsx
```diff
@@ -0,0 +1,50 @@
...
+  htmlType,
+  onClick,
+  children,
+  disabled,
+  style,
+  className,
+}: ButtonProps) {
+  return (
+    <ButtonStyles>
+      {/* eslint-disable-next-line */}
```
paschalidi, line 28: If you remove this you will see typescript gives you a very descriptive message on how to annotate the `htmlType`. Please check it out and when this is solved please remove this line. Remember we want to avoid disabling eslint in the actual code 99.9 percent of the time.

https://github.com/chakra-ui/chakra-ui/blob/4ce9e3c0f6/packages/button/src/button.tsx

Check this library and you will also find how to annotate the functions ;)

packages/ui/src/stories/Button/button.style.ts
```diff
@@ -0,0 +1,87 @@
...
+  destructive: "#E63644",
+  affirmative: "#189F59",
+  warning: "#D97508",
+  white: "#FFFFFF",
+  destructiveDim: "#FFE1E1",
+  affirmativeDim: "#E8F5E5",
+  warningDim: "#FFF6D1",
+  textInactive: "#BFBFBF",
+  background: "#F2F3F4",
+};
```
paschalidi, line 22: As you said is better if we export this.

packages/ui/src/stories/Button/Button.tsx
```diff
@@ -0,0 +1,50 @@
+import React from "react";
```
paschalidi, line 1: This file is now `Button/Button.tsx` can we make it `Button/index.tsx`? 

I think you will see the reasons, but if not we can discuss it

paschalidi (commented):
Hey look at my comments before this review and also check the comments  I make here. 

I am adding this as a comment because I thing `Request changes` is a little bit agrressive for my style at least. Would be nice tho to look in all my comments ;)

packages/ui/src/stories/Button/Button.tsx
AdrianValdes: I deleted. My bad I missed that. I already practiced configuring prettier with eslint-config-airbnb-typescript so that will help me not those silly mistakes. I am also thinking about adding husky to save you work and make me be more clean and precise with my commits. Do you think that is a good idea? I have never used but I found it doing research on the eslint/prettier issue.

packages/ui/src/stories/Button/Button.stories.tsx
AdrianValdes: You are right. I deleted already. I didn't think about it because you used it in the web-components repo. I left it there because I didn't do enough research on what that could mean, it was incorrect of me to do so.
I am still getting the sense of eslint and how to deal with it. I was reading yesterday until 2am.  But of course I shouldn't have left code I don't recognize.

packages/ui/.storybook/main.js
AdrianValdes: Deleted already too. I added to experiment because I thought we may need it. I should have asked first since I don't know how to really use it.

packages/ui/src/stories/Button/Button.tsx
AdrianValdes: Already removed the eslint line, same mistake as with the other one.

Regarding the htmlType annotations. I have been reading and I already understand what it means, and I also understand this annotation:

> The expected type comes from property 'type' which is declared here on type 'DetailedHTMLProps<ButtonHTMLAttributes<HTMLButtonElement>, HTMLButtonElement>'

From what I got React exports an interface HTML Elements and this interface has its own types:

` interface ButtonHTMLAttributes<T> extends HTMLAttributes<T> {
        autoFocus?: boolean;
        disabled?: boolean;
        form?: string;
        formAction?: string;
        formEncType?: string;
        formMethod?: string;
        formNoValidate?: boolean;
        formTarget?: string;
        name?: string;
        type?: 'submit' | 'reset' | 'button';
        value?: string | ReadonlyArray<string> | number;
    }`
    
  That is why Typescript complains when I use a type different than  'submit' | 'reset' | 'button' | undefined. Is that right?
  
  I am not sure how to solve it though, because we don't want to use those types, we want to use `primary`, `secondary`, etc, is that right?
  
   For that reason I am not sure what to do, I suppose we shouldn't change React's interface.

packages/ui/src/stories/Button/button.style.ts
AdrianValdes: Done.

packages/ui/src/stories/Button/Button.tsx
AdrianValdes: Already done. I see the reasons. I mean, I see how it helps us with the imports, maybe there are more reasons I don't know about.

packages/ui/src/stories/Button/Button.tsx
AdrianValdes: More on annotation.
 I decided to follow the advice of  [this](https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/basic_type_example) article, it is a cheat-sheet from Netlify, which I believe is a trust able company, is it? 
 Because of that I used the type `(event: React.MouseEvent<HTMLButtonElement>) => void` for the `onClick` function.
 
 For the children I used `React.ReactNode` because according to them is the best one. Nevertheless, it is a bit weird to me that children from a button can be virtually anything. Is it right to use `React.ReactNode`? I suppose we want to include cases when we use tags like `<span>` or `<icon>`

AdrianValdes:
Hi @paschalidi, I made all the changes you request and I commit them and pushed them. I have a doubt regarding the workflow. 

In this case, because the changes were small, I made them all at once and then commit them. Is that consider normal? Or should commit them as I fix them one by one?

AdrianValdes:
On another thought. I realize I need to practice Typescript a lot, as I have used just one time. I started to read an book and started also a small project (a minesweeper game) to practice. For this project I configured prettier with estlint-airbnb rules, so I now have a better understanding of how that works. 
I still have questions, there is some unsolved compatibility issues because of new version of packages being bundle in the latest React version. I found solutions, but we would have to discuss them to see if you think they are good for the project.
I am also using `husky`, I like what it does.

paschalidi:
Good stuff. Sounds solid. We can discuss all your questions on Monday :) Will have another look on this PR sometime soon

packages/ui/.storybook/main.js
paschalidi: Well we might wanna use it but as you say we need to investigate if it is of any use for us first

paschalidi: I reckon this is not deleted tho, can you double check?

packages/ui/src/Colors/index.ts
```diff
@@ -0,0 +1,22 @@
+const colors = {
```
paschalidi, line 1: looks solid. I like that you renamed it 👍

packages/ui/src/Colors/index.ts
```diff
@@ -0,0 +1,22 @@
...
+  warning: "#D97508",
+  white: "#FFFFFF",
+  destructiveDim: "#FFE1E1",
+  affirmativeDim: "#E8F5E5",
+  warningDim: "#FFF6D1",
+  textInactive: "#BFBFBF",
+  background: "#F2F3F4",
+};
+
+export default colors;
```
paschalidi, line 22: Lets export this as const same as we do in the Button component.

packages/ui/src/stories/Button/button.style.ts
```diff
@@ -0,0 +1,68 @@
...
+    }
+    &:disabled {
+      color: ${Colors.white};
+      background-color: ${Colors.secondary};
+      border: 1px solid ${Colors.secondary};
+    }
+  }
+`;
+
+export default ButtonStyles;
```
paschalidi, line 68: Lets not export these as default. In larger project I find it problematic, because on the import any developer can use any name they would like to. In this case this wont be an issue since these are styles and are imported only once but lets for rule of thumb always export consts.

packages/ui/src/stories/Button/index.tsx
```diff
@@ -0,0 +1,49 @@
+import React from "react";
+
+import ButtonStyles from "./button.style";
+
+export interface ButtonProps {
+  type: "primary" | "secondary";
+  id: string;
+  htmlType: string;
```
paschalidi, line 8:
```suggestion
  htmlType: 'button' | 'submit' | 'reset'| undefined;
```

Have you tried that? It is better to be more specific here. This way in the future when a new developer tries to use this library and passes `htmlType` a string that is _not_ one of those string we define above they will see an error.

packages/ui/src/stories/Button/button.style.ts
```diff
@@ -0,0 +1,68 @@
+import styled from "styled-components";
+
```
paschalidi, line 2:
```suggestion
```

packages/ui/src/stories/Button/Button.stories.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
```
paschalidi, line 2:
```suggestion
```
These are really silly suggestions. But I do them anyway

paschalidi (commented):
Looks good, made some last comments

packages/ui/src/stories/Button/index.tsx
AdrianValdes: I tried `'button' | 'submit' | 'reset'| undefined` but in our Button.stories.tsx, we are using `.add("secondary", () => <Button type="secondary">Call to Action</Button>)` so the linter throws an error.

And if I do something like 'button' | 'submit' | 'reset'| undefined | 'primary' | 'secondary' the linter throw an error because what I mentioned before:

> The expected type comes from property 'type' which is declared here on type 'DetailedHTMLProps<ButtonHTMLAttributes, HTMLButtonElement>'

packages/ui/src/stories/Button/Button.stories.tsx
AdrianValdes: Yes, yes, silly but necessary :). I have been thinking a lot about your attention to detail and how I have to copy that way of coding, it is amazing and you are right of course. I have to learn to be so consistent and organized, to triple check everything, to look at every line with fresh eyes. Thanks for taking the time to point all that out, really.

packages/ui/src/stories/Button/index.tsx
paschalidi: wait there are two different props, one we call it `htmlType`, and one `type`, the comments we are having is about the `htmlType`. Could that be `secondary`?

packages/ui/src/stories/Button/button.style.ts
AdrianValdes: I understand, I will change it. We will have to configure our linter rules once we add eslint-airbnb because once you added they advice to use `default export` in files with just one export declaration.

packages/ui/src/stories/Button/index.tsx
AdrianValdes: Oooooh, I see. So in the `Interface` we could have:
`type: "primary" | "secondary";
  htmlType: "button" | "submit" | "reset" | undefined;`
  
  And then:
  
  `<button
        type={htmlType}
        {...}
        </button>`

AdrianValdes: I really have to catch up with Typescript and Typescript in React.
Unfortunately I have German classes today, 6 hours!

AdrianValdes:
Once we merge this one I will start with the other task in the project config `adds eslint and prettier` (I hope it goes well because I set it up in another project already, but it looks that there is a lot of discussion  happening in the community because of conflicts with the new React version).

packages/ui/src/stories/Button/Button.stories.tsx
paschalidi: Think of eslint as a set of rules people set up in their projects that warns you about best practises.

packages/ui/.storybook/main.js
paschalidi: @AdrianValdes check this comment out please ☝️

paschalidi (approved):
Yop good work. Check one comment out and once you fix it you can merge this!

packages/ui/.storybook/main.js
AdrianValdes: Oh, for something reason I was not able to see this comment here, weird.

AdrianValdes: So should I delete every addon? I thought you refer just to addon-controls, my bad

packages/ui/src/stories/Button/button.style.ts
paschalidi: Yep right. Some rules we will change. But also we want to keep the changed rules to a minimum.

packages/ui/.storybook/main.js
AdrianValdes: I will delete these ones then:
` "@storybook/addon-links",
    "@storybook/addon-essentials",
    "@storybook/preset-create-react-app"
    `
    They were there by default.

paschalidi: Nope wait. Lets delete the one is added on this PR only and we can make a separate issue and discuss there possible addons that make sense.

Well I am saying this because you want in general to have a PR making one thing. Now if one PR makes multiple things then it starts being confusing for the reviewer and in general it isnt the best practise to mix things. 

Imagine a scenario that you merge a PR that by mistake breaks the application. In other words in pushes a bug in the application. In a real world app we want basically to be able to revert the changes pushed by that branch and _only_ those changes. Now if are used to have a PR does multiple things in scenarios like the one I am explaining you will end up not knowing what part of the changes broke the PR. 

This is why we keep it short and focused on one thing

packages/ui/src/stories/Button/Button.tsx
paschalidi: Exactly that!

packages/ui/src/stories/Button/Button.tsx
paschalidi: you are 100 percent right on your research! Good work :)

packages/ui/.storybook/main.js
AdrianValdes: That makes sense, so we just deleted ` "@storybook/addon-controls",` that was the one I removed already.

AdrianValdes: The others are there by default, so we will probably need them I suppose. Once I know more about storybook I will be able to have an opinion :)

paschalidi: Exactly right!

AdrianValdes: Perfect, I will merge then!

paschalidi, line 7: @AdrianValdes :) This was never deleted in the end

AdrianValdes: But I do deleted `"@storybook/addon-controls"`, and I am checking my git and it is all synchronized

AdrianValdes: I think I am understanding something wrong, sorry :(

AdrianValdes: I thought we agreed on not deleting 
` "@storybook/addon-links",
    "@storybook/addon-essentials",
    "@storybook/preset-create-react-app",`

paschalidi: Yep maybe its me. We dont want to delete the default but only the
```
"@storybook/addon-controls",
```suggestion
```
```

AdrianValdes:
I am checking the merged version, I mean, I am checking main and it is not
there. Do you see it? I am going to go crazy :) :)

AdrianValdes:
I am blind, sorry man! I am literally blind, well not literally but
shiiiiiiittt

paschalidi:
No worries just open a PR to fix it. These things are part of the process

AdrianValdes:
Now that we confirmed I am a blind person who is going to make YOU crazy,
what do I do? Should I create a new pull request and delete it?

AdrianValdes:
Ok, I got my answer

AdrianValdes:
I just deleted form the devDepencies, that was my mistake the whole time,
aaaaaah. I am sorry one more time. Thanks for your patience 💖

PR #18, opened 2021-01-27
fix: :bug: button types

AdrianValdes:


AdrianValdes:
In case you feel like reading some of these days, [this](https://dev.to/bytebodger/default-props-in-react-typescript-2o5o) article talks a lot about the default props issue. The author rejects the idea of using an interface and default function parameters, that threw me of too. Although now that I double check his example is slightly different.

AdrianValdes:
I pushed a lot yesterday morning or before, but I forgot to put you as a reviewer. I am going to kill myself! To be like that makes me crazy. I saw you change the title, it makes sense not to repeat fix.

paschalidi (approved):
Good stuff! Lets move into the gh actions next :)

paschalidi:
Want to delete your branches ;) @AdrianValdes 
image

AdrianValdes:
All of them?

paschalidi:
The ones that are merged. Dont delete them locally from your machine but from github

PR #21, opened 2021-01-29
refactor: renames folder and creates constants 

paschalidi:
hi :wave". I renamed the `stories` file into `components` and turned the `Colors` directory into one names `constants`. `constants` contains the colors we having now but in the future things can be added.


AdrianValdes (approved):
well thought! Merge it in!

PR #26, opened 2021-01-31
refactor: renames to small-ads

paschalidi:
I find small advertisements tedious to type. I renames the whole project to small-ads instead. Also did some cleanup and renaming of the packages.


AdrianValdes (approved):
Good idea! Merge it in!

PR #27, opened 2021-02-01
chore: improves scripts on the root package.json

paschalidi:
Hey 👋  I am updating those scripts on the root package. 

No you can run the apps from the root directory without having to navigate into different folders.

Check it out and let me know what you think buddy


AdrianValdes (approved):
Looks great!. Great stuff adding --stream and --parallel from `lerna`! I didn't know it was better that way.

PR #29, opened 2021-02-02
Add async handler

AdrianValdes:
Hi @paschalidi, 
this is still work in progress. I have a lot of doubts when using Typescript with `redux-toolkit`


src/webapp/src/characters/charactersSlice.ts
```diff
@@ -0,0 +1,83 @@
...
+      if (action.payload) {
+        return { ...state, error: action.payload.error };
+      }
+      return { ...state, error: action.error.message };
+    });
+
+    builder.addCase(getCharacters.fulfilled, (state, { payload }) => ({
+      ...state,
+      results: payload.results,
+      info: payload.info,
```
paschalidi, line 78: Why we store the info if we _never_ use it? We would always like to store only information we really need. Otherwise it can get confusing for other developers when they join a project or when they try to read the code, or solve a bug another developer has introduced.

src/webapp/src/characters/charactersSlice.ts
```diff
@@ -0,0 +1,83 @@
...
+    builder.addCase(getCharacters.rejected, (state, action) => {
+      if (action.payload) {
+        return { ...state, error: action.payload.error };
+      }
+      return { ...state, error: action.error.message };
+    });
+
+    builder.addCase(getCharacters.fulfilled, (state, { payload }) => ({
+      ...state,
+      results: payload.results,
```
paschalidi, line 77: In the same way, as the other comment, if useful information from this `results` object is only `{ name, species, image, id }` then we want to store in redux or anywhere, only those. The rest is noise we dont need :)

src/webapp/src/characters/charactersSlice.ts
```diff
@@ -0,0 +1,83 @@
...
+  { rejectValue: ApiKnownError }
+>('characters/getCharacters', async (_, { rejectWithValue }) => {
+  const response = await fetch(`https://rickandmortyapi.com/api/character`);
+
+  if (response.status === 404) {
+    return rejectWithValue((await response.json()) as ApiKnownError);
+  }
+
+  return (await response.json()) as ApiData;
+});
```
paschalidi, line 59: now what is you have a 500 error does that mean the promise has fulfilled? Most likely no. 

Try using the `response.ok` to check if the response if okay and in _any_ other case we can reject.

src/webapp/src/characters/charactersSlice.ts
```diff
@@ -0,0 +1,83 @@
...
+  initialState,
+  reducers: {},
+  extraReducers: (builder) => {
+    builder.addCase(getCharacters.pending, () => {});
+
+    builder.addCase(getCharacters.rejected, (state, action) => {
+      if (action.payload) {
+        return { ...state, error: action.payload.error };
+      }
+      return { ...state, error: action.error.message };
```
paschalidi, line 72: Mind explaining to me, when is the case that the code comes down here? I got a little bit confused

src/webapp/src/reducers.ts
```diff
@@ -1,7 +1,10 @@
...
 
-export const rootReducer = {
+export const rootReducer = combineReducers({
   person: personReducer,
-};
+  characters: charactersReducer,
+});
 
-export type RootState = typeof rootReducer;
+export type RootState = ReturnType<typeof rootReducer>;
```
paschalidi, line 10: 👍

paschalidi (commented):
Made quite some few comments please take a look and ask me if there is any questions 👍

src/webapp/src/characters/Characters.tsx
```diff
@@ -0,0 +1,39 @@
...
+    <div key={id} className='card'>
+      <div className='picture-wrap'>
+        <img src={image} title={name} alt='character' />
+      </div>
+      <div className='card-footer'>
+        <h3 className='character-name'>{name}</h3>
+        <div>{species}</div>
+      </div>
+    </div>
+  ));
```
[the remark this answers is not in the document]
AdrianValdes: Hi @paschalidi, I was specially looking forward to listen to your opinion in this topic. I have found a lot of different opinions and I am still trying to decided what _I_ personally find more readable, like you said, it is a very opinionated matter. 

Having said that, I made this abstraction thinking of you (that didn't work :)) because I am also inclined to think it is better to put the code directly in the `return` statement. So I changed it and I am fine with following this pattern as long as it makes sense. If not, I will make it a new component like you suggested (I just thought that in this case making a new component was unnecessary)

src/webapp/src/characters/Characters.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+import './Characters.css';
+
+export const Characters = () => {
+  const dispatch = useDispatch();
+  const { results, error } = useSelector(
+    (state: RootState) => state.characters
+  );
+
+  const charactersGrid = results.map(({ name, species, image, id }) => (
```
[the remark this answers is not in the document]
AdrianValdes: I see your point, for the variable makes more sense, specially if it is inside a `div` with a `characters-grid` `className`. I delete the variable anyways. 

Speaking of which, is the naming of the class also a bit off? Maybe `cards-collection` of `cards-grid` would be better? I would like to know what do you think it is better in general for naming cases like this one.

src/webapp/src/characters/Characters.css
```diff
@@ -0,0 +1,29 @@
...
+  width: 100%;
+}
+
+.card-footer {
+  padding: 1.5rem;
+}
+
+.character-name {
+  margin: 0;
+}
```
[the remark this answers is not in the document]
AdrianValdes: Regarding this point I have a question I cannot answer without your help I believe.

We discussed and agreed to use `styled-components` yes, but I thought we were going to use those components through `storybooks` and for now we cannot. So in order to use `stlyed-components` I have to install them as a dependency in the webapp.

Because of the above, I thought I would create just some dummy css. Nevertheless, if this was a bad decision I though of two ways of solving it.

1. I could learn how to use and build the components with `storybooks` and then use them in our webbapp
2. I could install `styled-components` in the `webapp` and learn how to use them for styling in this particular case.

What do you think it is better in cases like this one?

src/webapp/src/characters/Characters.tsx
```diff
@@ -0,0 +1,39 @@
...
+  return (
+    <div>
+      <h1>Characters</h1>
+      <button type='button' onClick={() => dispatch(getCharacters())}>
+        fetch
+      </button>
+      {error ? (
+        <p>An error has ocurred: {error}. Try again please</p>
+      ) : (
+        <div className='characters-grid'>{charactersGrid}</div>
```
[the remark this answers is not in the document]
AdrianValdes: Completely true! Bad programming, bad organization. Thanks again for pointing those things, I will get better at that.

src/webapp/src/characters/Characters.tsx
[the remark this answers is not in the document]
paschalidi: >(I just thought that in this case making a new component was unnecessary) 

It might be yes. Sometimes we make components not because they are reusable but to hide complexity. Like if you have a very complicated component that in the end displays some sort of list, then you might wanna create a component with a descriptive name. Then you will hide the complexity from the reader. Now if the reader is interested in that part they will open the component and start reading it. If not they will move on. 

Not this isnt very complex here, imo.

src/webapp/src/characters/Characters.tsx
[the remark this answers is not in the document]
paschalidi: I dont follow a good convention for css classes :P perhaps this is an opportunity to come up with one. `characters-grid` or `cards-grid` sounds legit.

src/webapp/src/characters/charactersSlice.ts
AdrianValdes: True, it makes sense and I agree. I had a reason to do that, but it was a wrong reason. Already solved.

src/webapp/src/characters/charactersSlice.ts
AdrianValdes: Also very true, I tend to store that kind of info *in case we need it* (crazy of me, I know!). I see how it is noise to have it there in the meantime. 

I removed everything we don't need. I like how the `store` looks now, much better. Thanks for pointing in it out.

src/webapp/src/characters/charactersSlice.ts
AdrianValdes: I see. I just checked against 404 because I used the `rejectWithValue()` function to reject with the specific error I know comes from the API in this case. Per the documentation: 

> If you are performing a request that you know will typically either be a success or have an expected error format, you can pass in a type to rejectValue and return rejectWithValue(knownPayload) in the action creator. This allows you to reference the error payload in the reducer as well as in a component after dispatching the createAsyncThunk action.

So my idea was to reject with the *known* error in case we have a 404.

In the other cases `createAsyncThunk` function is going to: 

> return a standard Redux thunk action creator. The thunk action creator function will have plain action creators for the pending, fulfilled, and rejected cases attached as nested fields.

So because `createAsyncThunk` automatically > Returns a fulfilled promise containing the final dispatched action (either the fulfilled or rejected action object).

I though to just handle the 404 with the `rejectWithValue` function and in the other cases let `creatAsyncThunk` pass the error to `getCharacters.rejected`

src/webapp/src/characters/charactersSlice.ts
AdrianValdes: Of course! Per the documentation:

> if the promise resolved with a rejectWithValue(value) return value, dispatch the rejected action with the value passed into action.payload and 'Rejected' as action.error.message

But:
> if the promise failed and was not handled with `rejectWithValue`, dispatch the rejected action with a serialized version of the error value as action.error

So if the promise failed and we passed the error to `rejectWithValue`, we can get that error from the `action.payload` and we
return `return { ...state, error: action.payload.error }`

But if the promise failed and we didn't rejected it using `rejectWithValue` `action.payload` is going to be `undefined` so we pass the default error that can be found in `action.error.message`

AdrianValdes:
hi @paschalidi :wave:! 

I made all the changes you suggested except for the using `response.ok` to check if the response was successful. I didn't make this change because I am against of doing it, I trust your experience, of course! 

I didn't change it because I don't understand how that change would fix the code and I would like to discuss the topic a little more to understand it. I deliberately introduced different network errors in the function to see the difference and I cannot find it. This is very important and I feel I may be missing something. It is that fine?

AdrianValdes:
I realize I just wrote a huge paragraph in the extra comments of the commit itself, which is probably not good. I am very sorry if that is awful. Let me know and I won't do it again.

src/webapp/src/characters/Characters.css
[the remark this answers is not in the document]
paschalidi: We gonna have styled components in storybook to style the simplest of the components. But also we gonna use them in the `webapp`. So both 1 and 2 are right in this case.

src/webapp/src/characters/charactersSlice.ts
paschalidi: The explanation is good. I will have a look onto it and let you know.

src/webapp/src/characters/Characters.css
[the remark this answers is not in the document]
AdrianValdes: Perfect, I will install `styled-components` in the `webapp` and I will change the Characters.css to practice because I haven't used it before. Now I have to go to the dentist to get a root canal! I am bit afraid.

src/webapp/src/characters/charactersSlice.ts
```diff
@@ -0,0 +1,63 @@
...
+    return rejectWithValue((await response.json()) as ApiKnownError);
+  }
+  const { results } = await response.json();
+
+  return results.map(({ species, image, id, name }: StateCharacter) => ({
+    species,
+    image,
+    id,
+    name,
+  }));
```
paschalidi, line 39: Reversing those two will make us catch more errors and avoid surprises. 
```suggestion
  if (response.ok) {
    const { results } = await response.json();
    return results.map(({ species, image, id, name }: StateCharacter) => ({
      species,
      image,
      id,
      name,
    }));
  }
  
  return rejectWithValue((await response.json()) as ApiKnownError);
```

Can you give this a shot please?

src/webapp/src/characters/charactersSlice.ts
```diff
@@ -0,0 +1,63 @@
...
+  initialState,
+  reducers: {},
+  extraReducers: (builder) => {
+    builder.addCase(getCharacters.pending, () => {});
+
+    builder.addCase(getCharacters.rejected, (state, action) => {
+      if (action.payload) {
+        return { ...state, error: action.payload.error };
+      }
+      return { ...state, error: action.error.message };
```
paschalidi, line 53: Trying some typescript magic here. Checking if payload is there which means we will be having an error message stored on the payload. When `payload` is `undefined` the message is in the `action.error.message`. 

I think this one line might read easier. What you think?
```suggestion
        return { ...state, error: action.payload?.error || action.error?.message  };
```

src/webapp/src/characters/Characters.css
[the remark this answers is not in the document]
paschalidi: Hahha good luck!

paschalidi (commented):
Please check those comments of mine out!

src/webapp/src/characters/charactersSlice.ts
```diff
@@ -0,0 +1,63 @@
+import { createAsyncThunk, createSlice } from '@reduxjs/toolkit';
+
+interface StateData {
```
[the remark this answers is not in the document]
AdrianValdes: I will avoid using **data** as a name! You are totally right! When you mention those things it becomes so clear and obvious in my mind that I don't know what I was thinking whenever I choose those names. 

Thanks for pointing these little things with so much patience man. You cannot imagine how much I am learning in each of your reviews! 

Since we started doing the project, whenever I am doing tutoring or helping the other students in their projects I am a machine of copying your way of communicating, and your advice about coding! I always give you the credit, don't worry! :D

src/webapp/src/characters/charactersSlice.ts
```diff
@@ -0,0 +1,63 @@
...
+  results: StateCharacter[];
+  error?: string | null;
+}
+
+export interface StateCharacter {
```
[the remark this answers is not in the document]
AdrianValdes: Dito with the above, you are 100% right!

src/webapp/src/characters/charactersSlice.ts
```diff
@@ -0,0 +1,63 @@
...
+  species: string;
+  image: string;
+}
+interface ApiKnownError {
+  error: string;
+}
+const initialState = {
+  results: [] as StateCharacter[],
+  error: null,
+} as StateData;
```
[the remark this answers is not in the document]
AdrianValdes: Yes, yes, much better! I was looking to do this but I kept doing something wrong and the compiler was complaining. I need learn how to use Typescript properly in these cases. For now, I started to understand a little of the mistakes the compiler throws at me. Thanks man, really appreciate ti.

src/webapp/src/characters/charactersSlice.ts
```diff
@@ -0,0 +1,83 @@
...
+
+interface ApiKnownError {
+  error: string;
+}
+
+export const getCharacters = createAsyncThunk<
+  ApiData,
+  void,
+  { rejectValue: ApiKnownError }
+>('characters/getCharacters', async (_, { rejectWithValue }) => {
```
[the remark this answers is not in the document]
AdrianValdes: I didn't see this one before, weird, I am wondering if I am doing something incorrectly when I check the reviews.

I agree, `fetchCharacters` is better, more descriptive. By the way, I had `fetchCaracters` and then while I was reading some article someone used `get<Variable>` and I changed it :(. Anyways, I am glad I have your support in this. It makes it easier to see what is *right*.

src/webapp/src/characters/charactersSlice.ts
```diff
@@ -0,0 +1,63 @@
...
+    builder.addCase(getCharacters.rejected, (state, action) => {
+      if (action.payload) {
+        return { ...state, error: action.payload.error };
+      }
+      return { ...state, error: action.error.message };
+    });
+
+    builder.addCase(getCharacters.fulfilled, (state, { payload }) => ({
+      ...state,
+      results: payload,
```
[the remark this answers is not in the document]
AdrianValdes: Beautiful  @paschalidi! It makes a lot, a lot of sense! Already done!

src/webapp/src/characters/charactersSlice.ts
AdrianValdes: It makes sense!! I like the approach. Checking if the response is successful to do stuff, and in any other case return the rejection, yes, beautiful.

src/webapp/src/characters/charactersSlice.ts
AdrianValdes: Great, thanks! 
I wanted to use the new optional chaining operator `?.` since then I read about it. But I don't have it present enough. I like the one line more, yes. More readable for sure, and cleaner! (Although I suppose in coding clean and readable are synonyms 99.99% of the cases :thinking: )

> Trying some typescript magic here. 

This magic is now doable in Javascript too, or am I wrong? It is not that you implied, I don't even know why I am asking :D

src/webapp/src/characters/charactersSlice.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+  const { results } = await response.json();
+
+  return results.map(({ species, image, id, name }: StateCharacter) => ({
+    species,
+    image,
+    id,
+    name,
+  }));
+});
```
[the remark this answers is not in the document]
AdrianValdes: I didn't see this one either. Two comment were missed somehow. I mean, they were not there... Is that possible? :thinking: Maybe it didn't refresh? Hmm... I have to figure that out. I wouldn't like to miss thinks by mistake.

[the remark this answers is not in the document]
AdrianValdes: I thought one of the ideas of `redux-toolkit` was to put everything concerning one slice in the same file. With the new way of creating actions we don't need an actions file, or not? In the end we don't need to use the `createAction` or the `createReducer` functions, we can do it everything with `createSlice`

I thought that would be good because I saw like that in the `redux-toolkit` documentation. 

BTW, I think there are still missing comments. I think when I was at the dentist I saw you mentioned something about the file name for the actions, but I cannot find it anymore.

AdrianValdes:
Hi @paschalidi , I made all the changes you suggested, I think they made the code look a lot better, seriously!!

I also moved the `fetchCharacters` function to another file like you suggested, but I have my doubts in how to do that the best way possible. The name I chose it is not the best, but I don't want to call it `actions` or `actionCreators`  because I suppose we would usually create a lot of actions using the `createSlice` functionality so we will export those actions from the same file using `charactersSlice.actions`. Should I call it `thunk`? 
I could swear you mentioned something about it, but I wasn't not in my computer and I cannot find were it was in case you did it. Maybe I dreamed it, or the anesthesia from the root-canal fucked with my head!

I also had to create a folder `types` to put the interfaces there, if not I will get an `eslint` error: `dependency cycle detected import/no-cycle`.

src/webapp/src/characters/fetchCharacters.tsx
```diff
@@ -0,0 +1,21 @@
+import { createAsyncThunk } from '@reduxjs/toolkit';
+import { ApiKnownError, Character } from './types';
+
+export const fetchCharacters = createAsyncThunk<
+  Character[],
+  void,
+  { rejectValue: ApiKnownError }
+>('characters/getCharacters', async (_, { rejectWithValue }) => {
```
paschalidi, line 8:
```suggestion
>('characters/fetchCharacters', async (_, { rejectWithValue }) => {
```

src/webapp/src/characters/Characters.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import './Characters.css';
+import { CharacterCard } from './CharacterCard';
+
+export const Characters = () => {
+  const dispatch = useDispatch();
+  const {
+    characters: { list },
+    error,
+  } = useSelector((state: RootState) => state.characters);
```
paschalidi, line 14: didnt really mean exactly that when I said 

```
characters : {
  list: [...]
}
```

Now if you see carefully we are having a state that looks something like this
```
characters : {
  error: "...",
  characters : {
    list: [...]
  }
}
```

paschalidi (commented):
Hey last bit and we get this in

src/webapp/src/characters/fetchCharacters.tsx
AdrianValdes: Of course! I have to think about those things whenever I make the changes. I need a better process to review everything I touch and not make this silly mistakes. Thanks

src/webapp/src/characters/Characters.tsx
AdrianValdes: Makes sense what you are saying. Already changed.

AdrianValdes:
BTW, I haven't made the changes in the css to use `styled-components`. I know we are going to delete this code, but maybe it is better do it anyways, we discussed that possibility.  Just to be 100% sure, should I do it?

paschalidi:
Up to you in any case lets do it in another PR :) Cant wait to merge this in!

paschalidi (approved):
GOOD WORK MAN! 👍

PR #30, opened 2021-02-04
fix: error on the eslintrc files

paschalidi:
was getting an error when I was opening the eslint file. I got this fixed. Can you also test locally that everything works well on your side as well?


AdrianValdes (approved):
hi @paschalidi, I pulled the branch and tested it locally and all is still working correctly. I suppose that means: *merge* :D

PR #33, opened 2021-02-05
fix: error on linter

AdrianValdes:
hi @paschalidi, 
I made clean pull from main and I started to have errors with the `linter`. Very weird because I had try to solution locally before and it didn't complain. To solve it on my side I first delete all the `node_modules` and the `package-lock.json` and re-install the project again, but that didn't solve it. 

I had to configure my `parserOptions` on my `.eslintrc.js` like this:

```
 parserOptions: {
    project: './tsconfig.json',
    tsconfigRootDir: __dirname,
  },
```

I have been reading about this issue and apparently there are discrepancies between `yarn` and `npm` for this case, but I though because we are using `lerna` that wouldn't matter :thinking:. 

Could you try this solution locally? It is a bit different to what I had before that was giving you trouble.


AdrianValdes:
BTW, the re-install changed some versions of the in the `package-lock.json` I don't know if that has something to do with this issue we are having.

paschalidi:
Can we start using yarn?

paschalidi:
Usually it isnt the best practise to delete those package.json.lock files. But I see where you coming from

AdrianValdes:
> Can we start using yarn?

Sure we can, if you think that is good, by all means. I can install it and start using it.

AdrianValdes:
> Usually it isnt the best practise to delete those package.json.lock files. But I see where you coming from

Oh, didn't know that. One time you told me to do it to solve one issue we were having with Github Actions and I thought something similar could be happening. I am going to read about `package-lock.json` files and why to treat them with care.

paschalidi:
Hm let me think for a second. I dont like that we are deleting those package.lock.json files so often. 

Perhaps we dont have to use yarn but can you explain why we need to delete the package.lock.json here? 
I reckon this is really last resort..

AdrianValdes:
I didn't know it was a last resort procedure. I thought it was a way of "cleaning" the installs and overriding "errors" there could be. I won't do it again. I will read about how they work as soon as I am back from the groceries.

AdrianValdes:
I am also going to:
1. Install yarn.
2. Clone to repository in a new file.
3. Install it using yarn.

Is that the correct procedure, or am I missing something? There is something I should pay attention to or read?

paschalidi:
We dont really need to use yarn to be honest, but yeah lets give it a shot. 

Lets do it in this current PR. 
1. Install yarn globally.
2. remove all the package.lock.json.
3. configure lerna to use yarn
4. run `yarn` to isntall deps
5. run `lerna bootstrap` to install deps in the projects

AdrianValdes:
hi @paschalidi,
Finally done. I ran into a few problems, I had to downgrade the version of Node, if not I would get errors when doing the bootstrapping. Anyhow, it works now without errors locally. Could you check if it works on your side?

AdrianValdes:
hi @paschalidi , I have some doubts regarding this topic. From now on we should install everything using `yarn add <packageName>` is that right?

paschalidi:
right

paschalidi (approved):
:thu

AdrianValdes:
I am getting this error, I have tried to find a solution but I haven'T been able to do it.

> yarn add v1.22.10
[1/4] Resolving packages...
error An unexpected error occurred: "https://registry.yarnpkg.com/eslint-config-small-ads: Not found".

paschalidi:
Wait before you merge this. I am checking it

paschalidi:
can you try on your root directory to run 

```
yarn lerna add styled-components --scope @small-ads/webapp  
```

AdrianValdes:
That worked

AdrianValdes:
BTW, Should I merge in the mean time? Are thing working on your side locally?

paschalidi:
yep that works for me. lets merge it

paschalidi:
https://github.com/lerna/lerna/issues/1886
There is an open issue for this on lerna. To remove the way you suggested you will be having to do what the person who posted the issue says they do.

PR #35, opened 2021-02-07
feat: :construction_worker: attempt to use storybook components

AdrianValdes:
Hi @paschalidi, 
In order to solve the issue "export the components from the `ui`", I added `@small-ads/ui` as a dependency of the `webapp`, and then ran `yarn bootstrap`. Afterwards I tried to import the Button component doing: `import { Button } from '@small-ads/ui/src/components/Button/';`. But I am getting the following error:

`File was processed with these loaders:  * ./node_modules/@pmmmwh/react-refresh-webpack-plugin/loader/index.js You may need an additional loader to handle the result of these loaders.`

I have been trying to solve it the whole morning but I haven't had any luck. There is no so much information that targets our specific setup. Could you give me some pointers?


AdrianValdes:
I finished the *Design Systems with React & Storybook* and didn't help :(. I don't think it is very good to be honest. I learned a few things but the speaker doesn't seem very good.

paschalidi:
I will try to help you on this one. 

Well you might know already that when we install packages that other developers wrote we access them as following.

```
import { useFunction } from 'react'
```

Now `react` or any other library what they have to do in their side is to create  single entry point for their whole library. This is usually an `index.js` file. Also usually libraries either written in js or in typescript they first get compiled to a minified version and then are being published to npm or yarn. This is because their size will be smaller and also sometimes compiling means making the code executable in all environment. For example you might know that browsers dont always have the latest of all js features. 

Now when you compile as

paschalidi:
I will try to help you on this one. 

Well you might know already that when we install packages that other developers wrote we access them as following.

```
import { useFunction } from 'react'
```

Now `react` or any other library what they have to do in their side is to create  single entry point for their whole library. This is usually an `index.js` file. Also usually libraries either written in js or in typescript they first get transpiled to a minified version and then are being published to npm or yarn. This is because their size will be smaller and also sometimes transpiling means making the code executable in all environment. 

For example you might know that browsers dont always have the latest of all js features. Now when libraries get transpiled, we make sure that the code can be executed in all the different environments one might have, like the browser or node or es5 etc etc. In other words when you write `class` in js there might be some environments that they dont recognise class but transpiling the code will drop it to their "version" sort to say so that it can be understood by it.

Now with typescript we _always_ have to transpile it to simple `js`.

This is what we have to do here now. 

Our goal is that in the end we will be able to import as following.
```
import { Button } from '@small-ads/ui'
``` 

To transpile typescript to javascript typescript itself gives you a command. The command is the `tsc`.
What we want to do in this PR is 
- [ ] write this script that will transpile the `ui` into a javascript. 
- [ ] create a single entry point for the `ui` and set it in the `package.json` as `main`. This way you tell the package what is its entry point. As the do [here](https://github.com/NiGhTTraX/ts-monorepo/blob/master/packages/components/package.json#L4)

This template is doing that, https://github.com/NiGhTTraX/ts-monorepo. But basically _any_ ts library is doing the same process. The difference is that we dont want to publish the library on `npm` because we are having a monorepo. 

Also this whole process would have happened either we used lerna or not so it is not specific to our own setup but is how packages are being exposed to the world. 

I think what I would do, is do some research and understand how to transpile the `ui` package to javascript using `tsc`. When you succeed with this you will be seeing an output generated in your repository. This is the entry point I am talking about. 


Hope this helps a little.

AdrianValdes:
hi @paschalidi ,
As we agreed I am pushing the code with the errors. I am still looking to find answers on how to transpile correctly for this case. Thanks for offering to look into it and thanks again for the good explanation you gave me above.

paschalidi (commented):
hey one thing, the `dist` files we arent pushing on github 95 percent of the time. Some people do but I wouldnt say is what you wanna be doing. 

Usually we put the files in the `.gitignore`

paschalidi:
@AdrianValdes I added a working version. Very not clean push and there is some things that we still have to work. 
I am a bit busy these days so could do a more clean work here. You mind if you take this to the end sir?

AdrianValdes:
I am working on in it. Right now I am having troubles because I cannot run the script `"build:js": "babel src --out-dir lib --extensions \".ts,.tsx\" --source-maps inline"`. It says I do not have babel installed. I saw you didn't added as a dependency. I suppose it has something to do with Babel coming integrated with `storybook`. I am reading about the issue.

paschalidi:
I think this would be possible without babel. Please try first to go about it without using babel and if there is no way then we can add babel

paschalidi (approved):

PR #36, opened 2021-02-07
connect firebase

AdrianValdes:
hi @paschalidi,
I connected the webapp to firebase. For now it is linked to a project I created in my firebase account. I sent you an invitation to: paschalidi@outlook.com, but of course  we can change it if it is needed. 

I tried to use Typescript powers (using Enums for example) and I tried to use the most moderns `redux-toolkit` techniques, but both Typescript and `redux-toolkit` are still new to me. I am doing other small projects at the same to practice them.


paschalidi:
Can we finish this one first please Adrian https://github.com/notifications?query=reason%3Aparticipating?

src/webapp/src/characters/Characters.tsx
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+import { Button } from '@small-ads/ui';
```
paschalidi, line 2: I think changes in this file are a bit irrelevant on the scope of this pr. Would be cool if you take a moment and revert them is good practise to focus on one thing at a time on prs helps both you and the reviewers have a focused scope 😎

src/webapp/src/firebase.ts
```diff
@@ -0,0 +1,14 @@
+import firebase from 'firebase/app';
+import 'firebase/auth';
+
+firebase.initializeApp({
```
paschalidi, line 4: Ideally here we would like to pass the key as an environment variable. Can you read a little about github secrets?

paschalidi (commented):
Hey looks great, I can't stress how much you have improved since we met dude! Code looks clean and reasonable. Check my comments there are few. However I am doing this review from mobile so I haven't yet gotten the chance to review the reducer file as I would have liked. Will do tomorrow morning though!

src/webapp/src/auth/userSlice.ts
```diff
@@ -0,0 +1,39 @@
...
+  extraReducers: (builder) => {
+    builder.addCase(signInUser.pending, (state) => {
+      state.authStatus = AuthStatuses.loading;
+    });
+    builder.addCase(signInUser.rejected, (state, action) => {
+      state.error = action.payload?.message || action.error?.message;
+      state.authStatus = AuthStatuses.notLoggedIn;
+    });
+    builder.addCase(signInUser.fulfilled, (state, { payload }) => {
+      state.user = payload;
```
paschalidi, line 23: seeing this from laptop again, I think it will be better to have `uid`, `displayName` and `email` here instead of `user`

src/webapp/src/auth/userSlice.ts
```diff
@@ -0,0 +1,39 @@
+import { createSlice } from '@reduxjs/toolkit';
+import { AuthStatuses, Login } from './types';
+import { signInUser, signOutUser } from './asyncAuthActions';
+
+const initialState: Login = {
+  user: null,
+  error: null,
```
paschalidi, line 7: we still dont know these values. `null` sometimes mean we know the value and is `null` whereas undefined means we still dont know what value is a variable having.

src/webapp/src/auth/Auth.tsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import styled from 'styled-components';
+import { useDispatch, useSelector } from 'react-redux';
+import { AuthStatuses } from './types';
+import { RootState } from '../reducers';
+import { signInUser, signOutUser } from './asyncAuthActions';
+
+const Loader = styled.div`
```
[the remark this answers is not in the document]
AdrianValdes: I used that loader as a placeholder, I think we could make something better :D. Would you prefer if I don't make PRs with those placeholders? I didn't want to try to make a real loader because we haven't decide important styling points.

src/webapp/src/auth/Auth.tsx
```diff
@@ -0,0 +1,64 @@
...
+    <div>
+      {authStatus === AuthStatuses.loggedIn ? (
+        <p>
+          Welcome: {user?.displayName} your id is {user?.uid} and you are using
+          the following email: {user?.email}
+        </p>
+      ) : (
+        <p>Hello there, login please</p>
+      )}
+      <button type='button' onClick={() => dispatch(signInUser())}>
```
paschalidi, line 56: can we do here the same as we do bellow, abstract that to its own function?

src/webapp/src/auth/Auth.tsx
```diff
@@ -0,0 +1,64 @@
...
+        LogIn
+      </button>
+      <button type='button' onClick={() => handleLogout()}>
```
paschalidi, line 59:
```suggestion
      <button type='button' onClick={handleLogout}>
```

can you try this?

paschalidi (commented):
some more comments while I was re-reviewing this from laptop

src/webapp/src/auth/Auth.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  if (authStatus === AuthStatuses.loading) {
+    return <Loader />;
+  }
+
+  return (
+    <div>
+      {authStatus === AuthStatuses.loggedIn ? (
+        <p>
+          Welcome: {user?.displayName} your id is {user?.uid} and you are using
```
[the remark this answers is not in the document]
AdrianValdes: I was checking in case a user wouldn't have one of those properties. I suppose in case of the `uid` and the `email` it doesn't make sense because they are going to be there always, just `displayName` is optional when we have other sing up methods, which we don't have now. So I suppose it is all completely unnecessary. Thanks!

src/webapp/src/auth/Auth.tsx
AdrianValdes: Of course, unnecessary wrapping!

src/webapp/src/auth/Auth.tsx
AdrianValdes: Got it. Code consistency!

src/webapp/src/auth/asyncAuthActions.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { User } from './types';
+
+export const signInUser = createAsyncThunk<
+  User,
+  void,
+  { rejectValue: AuthError }
+>('user/sigInUser', async (_, { rejectWithValue }) => {
+  try {
+    const { user } = await auth.signInWithPopup(provider);
+    const { displayName, uid, email } = user!;
```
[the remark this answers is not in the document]
AdrianValdes: hi @paschalidi,
This is a more or less new feature from Typescript. It is called the "non-null assertion operator", we can use it as a post-fix to assert that the variable is not `null` or `undefined` in cases the type checker cannot conclude that.

 I used it here because the function `auth.signInWithPopup(provider)`  returns a value with a type `firebase.User | null` so I am telling the compiler I know user wont be `null` here.
 
 Did I explain it right? Or should I try again?

src/webapp/src/auth/userSlice.ts
AdrianValdes: I made them null because the return type from the `auth.signInWithPopup(provider)` function is `firebase.User | null` so the compiler was complaining about me using undefined. I found a solution. I think it looks better now. Let me know if you are not convinced. Whatever is cleaner I am open to use it! Thanks for your advice!

src/webapp/src/auth/types.ts
```diff
@@ -1,16 +1,19 @@
...
   loggedIn,
   loading,
 }
 
 export interface Login {
-  user: User | null;
+  uid: string | undefined;
+  email: string | undefined;
+  displayName: string | undefined;
   error?: string | null;
```
paschalidi, line 17: Can we also have this as undefined?
```suggestion
  error?: string | undefined;
```

src/webapp/src/auth/asyncAuthActions.tsx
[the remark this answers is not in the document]
paschalidi: It sounds about right. Thanks!

src/webapp/src/auth/asyncAuthActions.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { UserData } from './types';
+
+export const signInUser = createAsyncThunk<
+  UserData,
+  void,
+  { rejectValue: AuthError }
+>('user/sigInUser', async (_, { rejectWithValue }) => {
+  try {
+    const { user } = await auth.signInWithPopup(provider)!;
+    const { displayName, uid, email } = user!;
```
paschalidi, line 13: Can we do that?
```suggestion
    const { user: { displayName, uid, email } } = await auth.signInWithPopup(provider)!;
```

paschalidi (commented):
Okay last round. Looks awesome!

src/webapp/src/auth/types.ts
AdrianValdes: Oh, I thought `null` was the correct value to use because we know we don't have an error at the beginning. I am going to ask you a little more about it later :ghost:

paschalidi: Right! You right that makes sense

src/webapp/src/auth/asyncAuthActions.tsx
AdrianValdes: hi @paschalidi, 
I wanted to do it like that but I couldn't! Without asserting that `user` is not null by doing `user!` the compiler complains.

paschalidi (approved):
Lets get this merged! Good work on this one dude. Lets spent time tonight figuring out next steps!

AdrianValdes:
I makes me really happy to hear that. Also when you said it at the beginning, you improved my mood. Thanks for all the effort and the support my friend. You have a beautiful soul, indeed.

PR #38, opened 2021-02-24
ci: remove bootstraping on Github actions

AdrianValdes:


AdrianValdes:
hi @paschalidi :wave: !
So far in this PR I have done 3 of the 4 required points for the linked issue.
1. I removed bootstrap (as we don't needed with the new `lerna` configuration we are using)
2. I added caching on the lint job of the `pr.yml`.
3. I added a new job on the `pr.yml`.
  3. 1 Should I add caching for this job too?
  3. 2 It is ok to put this job (prod-install) as part of the `pr.yml` or should we have a completely new `.yml` for that? I created a `production.yml` but I deleted because I thought it was not necessary
  
Now I want to tackle the Github Action for typechecking. I have been reading to find information about this topic, but I realized that I have not a clear picture of what we want to accomplish with this step. 

I have been thinking about these topics and I want to understand clearly: What is exactly typecheking in this case? Do we achieve that by transpiling?  Do we want to do it for the `ui` and the `webapp` apps?
 
Could you point me out to some documentation on that direction? I am going to read more about "tsc CLI Options" to try to find the answers I need.

paschalidi:
Did we discussed about including the `tsc` on the gh actions

AdrianValdes:
You told me I should add typechecking on GitHub Actions, at least that is what I understood. That is what I wrote and the issue but maybe I misunderstood something.

paschalidi:
Yep we did. Perhaps I should rephrase. I didnt see it in the code, is it being added?

AdrianValdes:
No not yet, I am not sure why we have to do it and I wanted to understand but I haven't been able of finding an explanation. I was also wondering if this is something that should be done both for the  `ui` and in the `webapp` apps?

AdrianValdes:
Maybe you know a good resource for me to check?

AdrianValdes:
If not I can just do it and we can talk about it in our meeting. If that is the case I am wondering if we are adding type checking for the `ui` and also the `webapp`?

.github/workflows/pr.yml
```diff
@@ -2,26 +2,41 @@ name: PR Workflow
...
-      run: yarn lint
-      
+      - uses: actions/checkout@v2
+      - name: Use Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: '15.x'
+
+      - name: Install dependencies
+        run: yarn install --prod
```
paschalidi, line 42: I think we can completely remove this bit. I think I didnt explain well what is the goal of this issue apologies. Lets forget what I told you about the production build. I would say its premature optimisation at this point

paschalidi (commented):
I will try to explain what we want to do here. 

Basically we want to add a script to both the `ui` and the `webapp` which will be typechecking. The script will be something very simple actually. In the package will be `"typecheck": "tsc"` or something similar. 

Then what we want is in addition to the lint script to also run this one.

Let me know if that makes sense the way I am describing it

.github/workflows/pr.yml
```diff
@@ -27,16 +27,6 @@ jobs:
...
-      - name: Use Node.js
-        uses: actions/setup-node@v2
-        with:
-          node-version: '15.x'
-
-      - name: Install dependencies
-        run: yarn install --prod
+      
+      -name: Typecheck
+        run: yarn run typecheck
```
paschalidi:
```suggestion
        run: yarn typecheck  
```

packages/ui/package.json
```diff
@@ -36,7 +36,7 @@
   "scripts": {
     "start": "start-storybook -p 6006 -s public",
     "build": "build-storybook -s public",
-    "build:js": "tsc -p tsconfig.build.json",
+    "typecheck": "tsc -p tsconfig.build.json",
```
paschalidi, line 39: add this extra one `"typecheck": "tsc"` and keep the `build:js` as it was. One is to build the package and one to do the typechecking.

.github/workflows/pr.yml
```diff
@@ -2,26 +2,31 @@ name: PR Workflow
...
+        if: steps.yarn-cache.outputs.cache-hit != 'true'
+        run: yarn install
 
-    - name: Lint
-      run: yarn lint
+      - name: Lint
+        run: yarn lint
       
+      - name: Typecheck
+        run: yarn run typecheck
```
paschalidi, line 32:
```suggestion
        run: yarn typecheck  
```

paschalidi (commented):
Check those comments :)

.github/workflows/pr.yml
AdrianValdes: > Let me know if that makes sense the way I am describing it

It makes sense. I added the typechking and the code runs without errors locally.

Nevertheless the Github Action is throwing this error again: `error TS2307: Cannot find module '@small-ads/ui' or its corresponding type declarations`. I am trying to solve it.

packages/ui/package.json
AdrianValdes: I did it, but when I ran the `"typecheck": "tsc"` the compiler created a lot of .js files besides the .ts files, I thought we wanted to avoid that. Let me try again.

AdrianValdes:
> And now we can use the action as single point of truth, so when it complains we will have to fix the PR first and then start fixing it. Hope that makes sense

I am lost. So from now on our single source of truth is the action I understand that : ). Then when an actions throws an error we have to fix the PR, I think that means we have to fix what is giving the error.
> and then start fixing it
 But what is it in this case?
 This is the part that confuses me: > so when it complains we will have to fix the PR first and then start fixing it.

AdrianValdes:
The code runs without errors locally.

Nevertheless the Github Action is throwing this error again: error TS2307: Cannot find module '@small-ads/ui' or its corresponding type declarations. I am trying to solve it.

.github/workflows/pr.yml
AdrianValdes: right, we just need `run` with `npm`

AdrianValdes:
> so when it complains we will have to fix the PR first and then start fixing it.

Oh, I think I got it. So any time the Github Action throws errors we first fix those errors and then we start cleaning and correcting the rest of the PR code. Is that it?

paschalidi:
Right so when I say start fixing it I meant fix this sort of errors  `error TS2307: Cannot find module '@small-ads/ui'` @AdrianValdes . Apologies for the confussion mate

AdrianValdes:
No worries, that is what I thought. I don't get the error: `error TS2307: Cannot find module '@small-ads/ui'` when I run `yarn typecheck` locally, just on the Github Actions. I am trying to find out why. Will keep you updated. Thanks!

AdrianValdes:
hi @paschalidi,
I solved the issue by implementing something you did sometime ago. We are now satisfying the compiler with on line `declare module '@small-ads/ui'`  inside `/src/types/@small-ads/ui/index.d.ts`. 
I am wondering if we should accept this or if there is something else I should do research on?

I also don't understand why we were not getting type-checking errors locally, but just on Github Actions.

paschalidi:
> I also don't understand why we were not getting type-checking errors locally, but just on Github Actions.

Probably you want to set up your environment again. 
Remember I had the same error locally last time we talked.

> I am wondering if we should accept this or if there is something else I should do research on?

This is what we would do if we were to write jasvascript. I believe the problem is again with the way we build the package. Remember we said we will merge that PR where the build was implemented even though my machine was throwing an error and yours didnt. We said then that we will implement the gh actions to see if there was an error really or not. If gh actions would prove we were having an error we would come back to the build process and find the best way to build the `ui` package.

AdrianValdes:
> The problem with using the one line solution is that we wont be having any types given to us when we actually using our components. That means we wont know what types our props are and so on which is why we use typescript in the first place.

I understand, it wouldn't make sense at all for us to do that! The thing is, I deleted all the node_modules and run `yarn` and I still don't get that error locally. 

I even have the types on the <Button> component. The Intellisense works and when hover over the Button I can see:

`(alias) function Button({ type, submit, disabled, style, className, children, onClick, }: ButtonProps): JSX.Element
import Button` 

I thought that means the types are working, or am I wrong?

> I believe the problem is again with the way we build the package.

Yes, you are probably right, I think I am missing something on the workflow. I will keep investigating. Just out of curiosity, do you have access to the types locally? Or still not?

paschalidi:
@AdrianValdes check this out

AdrianValdes:
hi @paschalidi ,
Already did :D, I even pushed in the middle of your process, sorry about that. I soon as I was reading what you did as I understand why we had the types locally but not on the Github Actions. It is almost obvious, we are not pushing the dist folder so we cannot have the types there!!

AdrianValdes:
At least I am happy I learned several things reading those files dozens of times and trying to understand what each property does.

paschalidi:
Its over :) I will check it and approve it tomorrow

paschalidi:
@AdrianValdes can you please try to get used to write description?
image

paschalidi (approved):
Lets get this in buddy! Good work and sorry for all the headaches you had to go through!

AdrianValdes:
> sorry for all the headaches you had to go through!

hi @paschalidi,
don't have to be sorry!! I am "suffering", that is correct :D, but I am also learning a lot. It is like taking some medicine, even if it doesn't taste super good, it is super important. So I am also happy you have the patience to wait for me to struggle enough before jumping to help.

AdrianValdes:
BTW, let me review the `firebase` PR again before you do it. I removed the requested review for now, will let you know soon.

PR #42, opened 2021-03-02
style: :construction: create Primary and Secondary buttons

AdrianValdes:
hi @paschalidi, 

Adrian here, so I am listened to your feedback and from now on I am going to try hard to maintain the lines of code per commit between 50-100. In this case I just crated the Primary and the Secondary buttons. 

1. I deleted the disabled property for the Secondary button because the Vercel people have just one disabled style.

2. Because we already had a `secondary` button in the `webapp` I decided to add also the `primary` one there. I think it is a little easier for you to see how they look. 
    2. 1 I am going to avoid doing this with more complicated components. But you mentioned it was fine to do it in some cases, 
    and I thought this was one of those cases. If that is wrong, please let me me know and I will remove it.

3. I didn't delete the rest of the colors from `color.ts` because I am trying to keep the commits small. I can do it afterwards if you think we should do it ASAP.

### DIFF
--- a/packages/ui/src/GlobalStyle/global.style.ts
+++ b/packages/ui/src/GlobalStyle/global.style.ts
@@ -5,13 +5,14 @@ export const GlobalStyle = createGlobalStyle`
     * {
         box-sizing: border-box;
     }
-    html {
-        font-size: 62.5%;
-    }
     body {
-        font-size: 1.6rem;
+        font-size: 1rem;
         font-family: "Inter", -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
         margin: 0;
         padding: 0;
+    }
+    button {
+        font-family: 'Inter', -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto,
+      Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
     }
 `;
--- a/packages/ui/src/GlobalStyle/global.style.ts
+++ b/packages/ui/src/GlobalStyle/global.style.ts
@@ -11,8 +11,4 @@ export const GlobalStyle = createGlobalStyle`
         margin: 0;
         padding: 0;
     }
-    button {
-        font-family: 'Inter', -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto,
-      Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
-    }
 `;
--- a/packages/ui/src/GlobalStyle/global.style.ts
+++ b/packages/ui/src/GlobalStyle/global.style.ts
@@ -1,0 +1,17 @@
+import { createGlobalStyle } from 'styled-components';
+
+export const GlobalStyle = createGlobalStyle`
+@import url('https://fonts.googleapis.com/css2?family=Inter:wght@300;500;600;700&display=swap');
+    * {
+        box-sizing: border-box;
+    }
+    html {
+        font-size: 62.5%;
+    }
+    body {
+        font-size: 1.6rem;
+        font-family: "Inter", -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
+        margin: 0;
+        padding: 0;
+    }
+`;
--- a/packages/ui/src/components/Button/button.style.ts
+++ b/packages/ui/src/components/Button/button.style.ts
@@ -6,60 +6,54 @@ export const ButtonStyles = styled.div`
   button:disabled .component-child {
     cursor: not-allowed !important;
   }
+  @import url('https://fonts.googleapis.com/css2?family=Inter:wght@300;500;600;700&display=swap');
+
   button {
-    font-size: 1.6rem;
-    display: flex;
+    height: 40px;
+    width: 140px;
+    min-width: auto;
+    max-width: calc(50vw - 36px);
+    display: inline-flex;
     align-items: center;
-    justify-content: space-evenly;
-    cursor: pointer;
-    font-weight: 400;
-    line-height: 1.9rem;
-    transition: background-color ease 0.2s;
-    height: 50px;
-    min-width: 200px;
-    box-shadow: rgba(0, 0, 0, 0.12) 0px 5px 10px 0px;
-    padding: 0;
-    border-radius: 5px;
+    justify-content: center;
+    font-size: 14px;
+    font-weight: 500;
+    font-family: 'Inter', --apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto,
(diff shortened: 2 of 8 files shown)


packages/ui/src/components/Button/button.style.ts
```diff
@@ -6,60 +6,53 @@ export const ButtonStyles = styled.div`
...
-    transition: background-color ease 0.2s;
-    height: 50px;
-    min-width: 200px;
-    box-shadow: rgba(0, 0, 0, 0.12) 0px 5px 10px 0px;
-    padding: 0;
-    border-radius: 5px;
+    justify-content: center;
+    font-size: 14px;
+    font-weight: 500;
+    font-family: 'Inter', --apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto,
```
paschalidi, line 21: I think we would like to have all our website having the same font. On how to do check this out
https://github.com/paschalidi/web-components/tree/master/src/GlobalStyles

packages/ui/src/components/Button/button.style.ts
```diff
@@ -6,60 +6,53 @@ export const ButtonStyles = styled.div`
   button:disabled .component-child {
     cursor: not-allowed !important;
   }
+  @import url('https://fonts.googleapis.com/css2?family=Inter:wght@300;500;600;700&display=swap');
```
paschalidi, line 9: Same as the other comment

paschalidi (approved):
Two comments the rest looks good!

packages/ui/src/GlobalStyle/global.style.ts
```diff
@@ -0,0 +1,17 @@
+import { createGlobalStyle } from 'styled-components';
+
+export const GlobalStyle = createGlobalStyle`
+@import url('https://fonts.googleapis.com/css2?family=Inter:wght@300;500;600;700&display=swap');
+    * {
+        box-sizing: border-box;
+    }
+    html {
+        font-size: 62.5%;
```
paschalidi, line 9: lets remove this, we dont really need it

packages/ui/src/GlobalStyle/global.style.ts
```diff
@@ -0,0 +1,17 @@
...
+    }
+    body {
+        font-size: 1.6rem;
```
paschalidi, line 12: lets make this 1rem

packages/ui/src/components/Button/button.style.ts
```diff
@@ -6,60 +6,50 @@ export const ButtonStyles = styled.div`
   button:disabled .component-child {
     cursor: not-allowed !important;
   }
+
   button {
-    font-size: 1.6rem;
-    display: flex;
+    height: 40px;
+    width: 140px;
```
paschalidi, line 12: lets make this min width

paschalidi (commented):
Just realised we said we want to have a button with the name `medium` and one with the name `large` perhaps we can pass this as props similar to what we do with the `type` of the button. 

Wanna take care of that here as well?

paschalidi (changes requested):
Changing the status of the PR

packages/ui/src/GlobalStyle/global.style.ts
AdrianValdes: I just can see you selected from line 1 to 9. Do you want me to remove line 5 to 10? 

We don't need the `box-sizing: border-box` anymore? I thought that will save us some trouble.

AdrianValdes:
> Just realised we said we want to have a button with the name `medium` and one with the name `large` perhaps we can pass this as props similar to what we do with the `type` of the button. 

Sure, that was my next step. I didn't do it now because I wanted to keep the PR small  :D

packages/ui/src/GlobalStyle/global.style.ts
paschalidi: I think I selected line 9 only. Didnt i? Now worries, I meant only `font-size: 62.5%;`

paschalidi:
Will approve then when we have these small changes changed

AdrianValdes:
I am also figuring out how to create the `customDecorator` to see our GlobalStyle component in storybook. That is why I didn't ask for a review, but you were too fast :D. 

I actually found out how to create a customDecorator, you did it [here](https://github.com/paschalidi/web-components/blob/master/.storybook/customDecorators.js). But I am trying to find the types for the storybook function, `storyFn` in this case.

packages/ui/src/GlobalStyle/global.style.ts
AdrianValdes: No problem, I will remove the 62.5% size then. I am also going to go back to using pixels for the button font-size then.

AdrianValdes:
hi @paschalidi,
I was out because we had a career workshop today at school, but I am back now :D

I have a question. While I have been studying a bit of Story book I came across [this](https://github.com/storybookjs/storybook/blob/master/lib/core/docs/storiesOf.md) article where they say `storiesOf()` is "Storybook's Legacy API". And although it is not officially deprecated yet, they claim they came with a better more portable solution with *Component Story Format ([CSF](https://storybook.js.org/docs/react/api/csf#gatsby-focus-wrapper))*. I thought that saying that we only have a button we could use the recommend the way? 

This is what they say:

> In Storybook 5.2 we introduced a simpler and more portable Component Story Format, and all future tools and infrastructure will be oriented towards CSF. Therefore, we recommend migrating your stories out of storiesOf API, and even provide automated tools to do this.

I am more than open to study the process and do refactor what it is necessary if something brakes. (Well, maybe something happens and I have to ask for your help but not until I cry of frustration :D

packages/ui/src/GlobalStyle/global.style.ts
```diff
@@ -5,13 +5,14 @@ export const GlobalStyle = createGlobalStyle`
...
-        font-size: 1.6rem;
+        font-size: 1rem;
         font-family: "Inter", -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
         margin: 0;
         padding: 0;
     }
+    button {
+        font-family: 'Inter', -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto,
+      Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
+    }
```
paschalidi, line 17: Are you sure we need this?

paschalidi:
> I thought we could use the recommend the way?

Yes good point lets do it

packages/ui/src/GlobalStyle/global.style.ts
AdrianValdes: If we don't declare the `font-family` for button as well, the user agent stylesheet overrides the font we have in the body because the specificity is bigger. Do you have another solution?

paschalidi: Perhaps try see if `inherit` will do the job?

AdrianValdes: Yep, very true `inherit` will do the job! My bad.

AdrianValdes:
> Yes good point lets do it

Ok, in the next PR then.

paschalidi (approved):
sweet work thanks dude!

PR #44, opened 2021-03-03
refactor: :construction_worker: use Component Story Format

AdrianValdes:
hi @paschalidi,
I migrated the code to use CSF, I think it looks better and it is easier to maintain the components now. It is also easier to see them on storybook, check how on can now change different properties directly in each view. 

1. I added two rules to the `.eslintrcjs` in the `ui`, I know we want to avoid it, but in this case I think it makes sense. Also, It is required to use export defaults to use CRF. In the end we are just modifying the rules for `ui` so it should be fine, I think. Let me know if I am missing something.

### DIFF
--- a/packages/ui/.eslintrc.js
+++ b/packages/ui/.eslintrc.js
@@ -7,5 +7,7 @@ module.exports = {
   ignorePatterns: ['dist/**/*'],
   rules: {
     'import/no-extraneous-dependencies': 'off',
+    'import/no-default-export': 'off',
+    'react/jsx-props-no-spreading': 'off',
   },
 };
--- a/packages/ui/.eslintrc.js
+++ b/packages/ui/.eslintrc.js
@@ -4,10 +4,8 @@ module.exports = {
     project: './tsconfig.json',
     tsconfigRootDir: __dirname,
   },
-  ignorePatterns: ['dist/**/*'],
+  ignorePatterns: ['dist/**/*', '*.stories.tsx'],
   rules: {
     'import/no-extraneous-dependencies': 'off',
-    'import/no-default-export': 'off',
-    'react/jsx-props-no-spreading': 'off',
   },
 };
--- a/packages/ui/.storybook/main.js
+++ b/packages/ui/.storybook/main.js
@@ -1,8 +1,9 @@
 module.exports = {
-  stories: ["../src/components/**/*.stories.(ts|tsx|js|jsx|mdx)"],
+  stories: ['../src/components/**/*.stories.(ts|tsx|js|jsx|mdx)'],
   addons: [
-    "@storybook/addon-links",
-    "@storybook/addon-essentials",
-    "@storybook/preset-create-react-app",
+    '@storybook/addon-links',
+    '@storybook/addon-essentials',
+    '@storybook/preset-create-react-app',
+    '@storybook/addon-actions',
   ],
 };
--- a/packages/ui/src/components/Button/Button.stories.tsx
+++ b/packages/ui/src/components/Button/Button.stories.tsx
@@ -6,7 +6,6 @@ const Template: Story<ButtonProps> = (args) => (
   <Button {...args}>Call to action</Button>
 );
 
-// 👇Each story then reuses that template
 export const Primary = Template.bind({});
 
 export const Secondary = Template.bind({});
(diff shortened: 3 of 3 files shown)


AdrianValdes:
I have been trying to add the other sizes for the `button`. Now, I can do using `className`, but I believe you prefer to do something similar to what we did with the `type` property.

 I having trying to do it for a while now, but I am having troubles to do it because the property 'size' does not exist on type 'DetailedHTMLProps<ButtonHTMLAttributes<HTMLButtonElement>,HTMLButtonElement>'. I am trying to extend our interface for `ButtonProps`, but others problems appear. It is driving me crazy! I will drop you a line if it gets to a point where I hate everything around me :D

paschalidi:
> I having trying to do it for a while now, but I am having troubles to do it because the property 'size' does not exist on type 'DetailedHTMLProps<ButtonHTMLAttributes,HTMLButtonElement>'. I am trying to extend our interface for ButtonProps, but others problems appear. It is driving me crazy! I will drop you a line if it gets to a point where I hate everything around me :


Wait I think I didnt exlain myself well. You will indeed use classes for this. But when we use the `Button` it will be something like 
```
<Button size="medium" />
```

Inside the compomnent this prop will change the height using classes

AdrianValdes:
> Inside the compomnent this prop will change the height using classes.

You explained yourself well, I think I am the one who is explaining it wrong. When we try to put `size` as a property of a button like so:
```
<Button size="medium" />
´´´

The compiler complains because `size` is not a native property of the buttons in HTML. [Here](https://github.com/typescript-cheatsheets/react/pull/276) is the issue.

File: packages/ui/.eslintrc.js [hunk 1]
```diff
@@ -7,5 +7,7 @@ module.exports = {
   ignorePatterns: ['dist/**/*'],
```
paschalidi, line 7:
```suggestion
  ignorePatterns: ["dist/**/*", "*.stories.tsx"],```

File: packages/ui/.eslintrc.js [hunk 2]
```diff
@@ -7,5 +7,7 @@ module.exports = {
...
   rules: {
     'import/no-extraneous-dependencies': 'off',
+    'import/no-default-export': 'off',
+    'react/jsx-props-no-spreading': 'off',
```
paschalidi, line 11:
```suggestion
```

Let add the stories into the ignore pattern instead.
I prefer having clean components that they follow the same patters and leave the stories flexible to use all the techniques we want. Stories is not code users will use in production in the end

packages/ui/.storybook/main.js
```diff
@@ -1,8 +1,9 @@
...
-  stories: ["../src/components/**/*.stories.(ts|tsx|js|jsx|mdx)"],
+  stories: ['../src/components/**/*.stories.(ts|tsx|js|jsx|mdx)'],
   addons: [
-    "@storybook/addon-links",
-    "@storybook/addon-essentials",
-    "@storybook/preset-create-react-app",
+    '@storybook/addon-links',
+    '@storybook/addon-essentials',
+    '@storybook/preset-create-react-app',
+    '@storybook/addon-actions',
```
paschalidi, line 7: Looks good!

packages/ui/src/components/Button/Button.stories.tsx
```diff
@@ -1,39 +1,28 @@
...
-  .add('submit', () => (
-    <Button onClick={() => {}} submit>
-      Call to Action
-    </Button>
-  ));
+const Template: Story<ButtonProps> = (args) => (
+  <Button {...args}>Call to action</Button>
+);
+
+// 👇Each story then reuses that template
```
paschalidi, line 9: Thanks!

paschalidi (commented):
Looks great! Thanks for the small PRs <3 One comment only

packages/ui/.eslintrc.js
AdrianValdes: Great solution! I keep forgetting about the `ignorePatters`! I hope this is the last time I forget about it! Thanks man

paschalidi (approved):
SUPER

AdrianValdes:
> Inside the compomnent this prop will change the height using classes

I got it! I was trying to do that but with an unnecessary work around. It took me a while :( Thanks mate

Issue #45, opened 2021-03-04
All components can live inside the `components` folder


paschalidi:
Even though `GlobalStyles` is only css it is a react component. Can we move it inside the components folder?
![image](https://user-images.githubusercontent.com/4181674/109931246-84e38380-7cc0-11eb-8ffb-091bdee8446d.png)

PR #46, opened 2021-03-04
chore: move `globalStyle` to components folder

AdrianValdes:


AdrianValdes:
Hi @paschalidi,
You are right about this, in the end `globalStyle` is also a component!

P.S: This is PR is very small. I can add the sizes here if you think it makes sense. But maybe not so much because the PR is linked to the All components can live inside the `components` folder #45  issue.

paschalidi (approved):
Small is good

PR #47, opened 2021-03-04
feat: add small and medium variants to the Button

AdrianValdes:


paschalidi (commented):
Code looks good. 

Can we add two more stories, one medium secondary button and one small secondary button. I know you can use the controls to do so but I think would be cool to have it anywas

packages/ui/src/components/Button/Button.stories.tsx
```diff
@@ -11,11 +11,17 @@ export const Primary = Template.bind({});
 export const Secondary = Template.bind({});
 Secondary.args = { type: 'secondary' };
 
-export const Medium = Template.bind({});
-Medium.args = { size: 'medium' };
+export const MediumPrimary = Template.bind({});
+MediumPrimary.args = { size: 'medium' };
```
paschalidi, line 15: Not so sure we are using this right. 
```suggestion
export const MediumPrimary = <Button  size="medium">Call to action</Button>
```

isnt this the same?

AdrianValdes: > Not so sure we are using this right.

I am trying to use what they recommend [here](https://storybook.js.org/docs/react/writing-stories/args). There is a few advantages when using `args`

paschalidi: Right okay I see thanks for explaining

paschalidi (approved):
Lets get this merged

packages/ui/src/components/Button/Button.stories.tsx
AdrianValdes: > isnt this the same?

Well in this case it will throw errors. To go over them we would have to first  to add `onClick={() => {}}` to every Button. And then stop using `args`. And `args` it what allows Storybook and its addons to live edit components.

PR #48, opened 2021-03-04
chore: make `typescript` a devDependency

AdrianValdes:
hi @paschalidi,
Even though the issue says

> Make typescript devDependency in all the repos we have

I didn't do it on @small-ads because I thought there we do need as a dependency. Am I wrong?
Sorry I ask, I am trying to get see what parts of how the process works I am missing.

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -21,7 +21,6 @@
     "react-dom": "^17.0.1",
     "react-scripts": "4.0.1",
     "styled-components": "^5.2.1",
-    "typescript": "^4.1.3",
     "web-vitals": "^0.2.4"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


AdrianValdes:
hi @paschalidi, 
Even though the issue says 

> Make typescript devDependency in all the repos we have

 I didn't do it on `@small-ads` because I thought there we do need as a dependency. Am I wrong?

paschalidi (commented):
Right, what about type like for instance `@types/styled-components` do these need to be dependencies?

paschalidi:
> I didn't do it on @small-ads because I thought there we do need as a dependency. Am I wrong?

No need to be sorry for asking dude. You right for the eslint package we need typescript as dependency

AdrianValdes:
I couldn't move some of them because the linter complains: 

> '@firebase/auth-types' should be listed in the project's dependencies, not devDependencies.

paschalidi (approved):
thanks looks good

PR #50, opened 2021-03-04
feat: add routing to Login page

AdrianValdes:


src/webapp/src/Navbar.tsx
```diff
@@ -20,12 +21,20 @@ const Header = styled.div`
...
-      <Button size='medium' onClick={() => {}}>
-        Login
-      </Button>
-    </Header>
-  </HeaderStyles>
-);
+export const Navbar = () => {
+  const history = useHistory();
+  return (
+    <HeaderStyles>
```
paschalidi, line 27: Looks like there is a border missing on the bottom of the navbar. Mind taking a moment in this PR to add it?
image

src/webapp/src/Navbar.tsx
```diff
@@ -20,12 +21,20 @@ const Header = styled.div`
...
+      <Header>
+        <Button
```
paschalidi, line 29: Also we want this button to be disabled when the user is in the login page

src/webapp/src/auth/Login.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+export const Login = () => (
```
paschalidi, line 3: These are the authentication options page to me. Perhaps we can call it `AuthenticationOptions`?

src/webapp/src/App.tsx
```diff
@@ -1,22 +1,31 @@
...
+      <Route exact path='/'>
+        <Auth />
+        <Person />
+        <Button onClick={() => {}} type='secondary'>
+          Invite Team
+        </Button>
+        <Button onClick={() => {}} type='primary'>
+          New Project
+        </Button>
+        <Characters />
```
paschalidi, line 27: Can we get rid of those in this PR please?

paschalidi (commented):
Overall looks good, minor comments only.

paschalidi:
Also lets put the Navbar in a new directory called `components`, what you think? In that file lets follow the same structure as the one in storybook. Aka `Navbar` folder which will contain styles tests and the component itself. What you think?
image

src/webapp/src/Navbar.tsx
AdrianValdes: The border appear and disappear in the home route depending on the position of the scroll bar. In the login routes is where is always there. I have to find a way of doing that nicely.

paschalidi: We can start by always having it there and find a way to remove it only in certain cases

src/webapp/src/App.tsx
AdrianValdes: I was hoping for you to say that! I hate there! :D

AdrianValdes: I should only remove them from the UI, right? 

Or do you want me to completely remove the components from the project as well?

AdrianValdes:
> Also lets put the Navbar in a new directory called components, what you think? In that file lets follow the same structure as the one in storybook. Aka Navbar folder which will contain styles tests and the component itself. What you think?

I like a lot this structure, I think it is going to help us to be organized. Thanks!

src/webapp/src/auth/Login.tsx
AdrianValdes: Done

src/webapp/src/App.tsx
paschalidi: If you think we might need them again lets keep them. But I doubt it. However feel free to remove them in another PR.

AdrianValdes: Yes, I will do it in another PR better.

PR #51, opened 2021-03-04
chore: builds the ui before starting the app

paschalidi:
hey adding this bit on the start:webapp script. Find it useful that the ui package is build for me when I start the app.


AdrianValdes (approved):
Good stuff, I was thinking about this yesterday! I remember I thought: "There has to be a way of building the package automatically, doing manually is not a very efficient way". Thanks for this dude!

PR #52, opened 2021-03-04
feat: add routing to Login page

AdrianValdes:
hi @paschalidi,
I had to create this branch because something broke in one commit and I didn't find a way to role it back. After trying to solve it for two hours I start doing research and I think I found a good solution (but of course I am not sure, first time I see something like this).

As you can see I went back to the last working commit, created a new branch, and redo everything again. I did step by step to see if I could find a mistake, but all tests passed successfully. What do you think? Is it horrible what I just did? I am looking at the Network graph in the insights and the branches look good.

### DIFF
--- a/src/webapp/src/components/Navbar/Navbar.tsx
+++ b/src/webapp/src/components/Navbar/Navbar.tsx
@@ -34,7 +34,7 @@ export const Navbar = () => {
             history.push('/login');
           }}
         >
-          Login Test
+          Login
         </Button>
       </Header>
     </HeaderStyles>
--- a/src/webapp/src/components/Navbar/Navbar.tsx
+++ b/src/webapp/src/components/Navbar/Navbar.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { Button } from '@small-ads/ui';
 import styled from 'styled-components';
-import { useHistory } from 'react-router-dom';
+import { useHistory, useRouteMatch } from 'react-router-dom';
 
 const HeaderStyles = styled.div`
   position: sticky;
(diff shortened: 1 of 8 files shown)


src/webapp/src/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,44 @@
...
+`;
+
+export const Navbar = () => {
+  const history = useHistory();
+  const route = useRouteMatch('/login');
+  return (
+    <HeaderStyles>
+      <Header>
+        <Button
+          disabled={!!route?.isExact}
```
paschalidi, line 33:
```suggestion
          disabled={Boolean(route?.isExact)}
```

This way we can always know this is a boolean. Less error prone, mostly you can see the difference in most complex code but I am bringing this to your attention now even tho this might not be that complex. No need to apply this change just so you know this is an alternative which is more close to english language

AdrianValdes: Great, great! Yes, please point out all those little things, I enjoy to learn them and I don't think it is annoying at all! I wanted to use something more robust than `!!` so I tried something with `typeof route?.isExact === undefined` but the linter complained.

paschalidi (approved):
Code looks great. I would hire you if I would see code like this in your code challenge ;)

AdrianValdes:
Thanks man :D! It means a lot to me, specially coming from you, you cannot imagine how much I am learning from all this, or maybe you do imagine it ;)

Nothing related to the commit at all: Today DCI asked me if I would be interested on becoming an assistant teacher there, I am going to think about it and I will talk to you more about it once I have more information.

PR #53, opened 2021-03-04
feat: add social button for Facebook

AdrianValdes:


AdrianValdes:
Hi @paschalidi , 
This is not the final design of the button, I made a PR thinking to see if you agree with this way of creating the social buttons in Storybook.

paschalidi (commented):
I think we want to give the user the tools to build these buttons if they would like so but not by passing a prop such as `social`. 
Let me explain myself. We now have buttons that are small and medium. Looks to me that this button is a large type of button. So 
1. we want to have the opportunity to use the Button component with a large size. 
2. also we want the user to be able to pass an icon to the button. This will be its own prop. And then the user will be passing in an icon if they would like so.
3. you want to make another prop that when is passed will make the button to take full width of its container.
4. the rest will be simply custom classes and styles we want to have on the Buttons. 

I hope you see how this way we create reusable components that can be generically used in other use cases as well

AdrianValdes:
> Looks to me that this button is a large type of button. 

I created an option for large now, it is the default size of our buttons. The same as Vercel.

AdrianValdes:
Hi @paschalidi ,
I think is a good idea to do everything you suggested. I am on it.

> you want to make another prop that when is passed will make the button to take full width of its container.

I am not 100% happy with the way I solved this, it is not terrible but there has to be a better way. Do you have any ideas?

AdrianValdes:
> 3. you want to make another prop that when is passed will make the button to take full width of its container.

I pushed another proposal. Still not 100% happy, I am debating with myself to decide what is cleaner of in which way I should move.

AdrianValdes:
Hi @paschalidi,

I think this PR may be too big, I stayed under 150 lines, but I believe it would be better if the PR would be smaller. I tried to dived it the process into smaller commits, but I don't know if that is really helpful. 

I pushed so many new lines because I was trying to achieve the 4 steps you gave me. Could you give some advice on what to do in cases like this one? How can I improve and make your/our life easier with this? 

Thanks for dividing the tasks today man, I had no idea of what to do some of those things so it was great to have those steps. And I definitely agree with you on this:  

> hope you see how this way we create reusable components that can be generically used in other use cases as well

paschalidi:
Well if in the PR we do only one thing which is basically adding a way to create these buttons then it isnt that big of a problem. 150 lines is a good number. For many cases this is okay. 

People might ask you to break the PRs into smaller ones only when you are trying to solve problems things at once. If this is not the case then I am happy to review.

packages/ui/src/components/Button/index.tsx
```diff
@@ -9,7 +9,9 @@ export interface ButtonProps {
   className?: string;
   onClick: (event: React.MouseEvent<HTMLButtonElement>) => void;
   children: React.ReactNode;
-  size?: 'small' | 'medium';
+  size?: 'small' | 'medium' | 'large' | '';
```
paschalidi, line 12: Mind taking a second and explain what makes us need the `''`? Cant we default to one of the sizes, perhaps medium?

packages/ui/src/components/Button/index.tsx
```diff
@@ -9,7 +9,9 @@ export interface ButtonProps {
...
+  full?: boolean;
```
paschalidi, line 13: Perhaps we call this
```suggestion
  fullWidth?: boolean;
```

packages/ui/package.json
```diff
@@ -19,6 +19,7 @@
     "react-dom": "^17.0.1",
     "react-scripts": "4.0.1",
     "styled-components": "^5.2.1",
+    "styled-icons": "^10.29.0",
```
paschalidi, line 22: If we use these only in the stories then this will be a `devDependency` dont you think?

src/webapp/src/auth/AuthenticationOptions.tsx
```diff
@@ -1,7 +1,68 @@
...
+  align-self: center;
+`;
+
+const GeistSpacer = styled.span`
+  margin-left: 23px;
+  margin-top: 23px;
+`;
+
+const FbButton = styled(Button)`
+  &&& {
```
paschalidi, line 40: Never seen this before, what does it do?

src/webapp/src/auth/AuthenticationOptions.tsx
```diff
@@ -1,7 +1,68 @@
+import { Button } from '@small-ads/ui';
 import React from 'react';
+import styled from 'styled-components';
+import { FacebookSquare } from '@styled-icons/boxicons-logos';
+
+const LoginScreen = styled.div`
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  height: calc(100vh - 64px);
```
paschalidi, line 10: The menu will always be 64px. Could we extract this to a constant?

AdrianValdes:
I was just trying to solve the four points you gave me. Well, that is not 100% true, I did made a very tiny change in the `main.js` from Storybook, we were using a [deprecated](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#correct-globs-in-mainjs) way I literally just added one **@*** for our blob. 

I am sorry for mixing that in the middle, next time I will create a separated issue and I will do it there. I understand now that even though is just a line is noise and more work for you. I will be more careful, my bad.

src/webapp/src/auth/AuthenticationOptions.tsx
```diff
@@ -1,7 +1,68 @@
...
 export const AuthenticationOptions = () => (
-  <div>
-    <h1>Login here</h1>
-  </div>
+  <LoginScreen>
+    <LoginMain>
+      <LoginSign>Login to Vercel</LoginSign>
+      <LoginForm>
+        <GeistSpacer aria-hidden='true' />
+        <FbButton
```
paschalidi, line 58: Better not get into the habit of using abbreviations.

packages/ui/src/components/Button/button.style.ts
```diff
@@ -59,10 +60,24 @@ export const ButtonStyles = styled.div`
...
+  .large {
+    height: 50px;
+    span {
+      position: absolute;
+      display: flex;
+      -webkit-box-align: center;
+      align-items: center;
+      inset: 0px auto 0px 22px;
+      z-index: 1;
+    }
```
paschalidi, line 78: I would expect we will use the button with an icon even when the size is small. Mind handling this case as well?

src/webapp/src/auth/AuthenticationOptions.tsx
```diff
@@ -1,7 +1,68 @@
...
+  }
+`;
 
 export const AuthenticationOptions = () => (
-  <div>
-    <h1>Login here</h1>
-  </div>
+  <LoginScreen>
+    <LoginMain>
+      <LoginSign>Login to Vercel</LoginSign>
```
paschalidi, line 55: Why we cal this Vercel? We only copy their style :P

src/webapp/src/auth/AuthenticationOptions.tsx
```diff
@@ -1,7 +1,68 @@
...
+`;
+
+const LoginForm = styled.div`
+  width: 100%;
+  text-align: center;
+  max-width: 320px;
+  align-self: center;
+`;
+
+const GeistSpacer = styled.span`
```
paschalidi, line 34: What `Geist` means for us?

src/webapp/src/auth/AuthenticationOptions.tsx
```diff
@@ -1,7 +1,68 @@
...
+const LoginMain = styled.div`
+  align-self: center;
+  flex: 1 1;
+  display: flex;
+  justify-content: center;
+  flex-direction: column;
+`;
+
+const LoginSign = styled.div`
+  font-size: 3rem;
```
paschalidi, line 22: Are we using px or rem? you said before you will use pxls no?

src/webapp/src/auth/AuthenticationOptions.tsx
```diff
@@ -1,7 +1,68 @@
...
+  letter-spacing: -0.066875rem;
```
paschalidi, line 23: That looks strange. What is the purpose it serves?

src/webapp/src/auth/AuthenticationOptions.tsx
```diff
@@ -1,7 +1,68 @@
...
+
+const LoginForm = styled.div`
+  width: 100%;
+  text-align: center;
+  max-width: 320px;
+  align-self: center;
+`;
+
+const GeistSpacer = styled.span`
+  margin-left: 23px;
```
paschalidi, line 35: Usually people use for padding derivatives of `4`. Its a widely used convention. Can we keep this rule, even if in vercel they dont.. Remind to talk to you about this.

src/webapp/src/auth/AuthenticationOptions.tsx
```diff
@@ -1,7 +1,68 @@
...
+      border: 1px solid #166fe5;
+    }
+  }
+`;
 
 export const AuthenticationOptions = () => (
-  <div>
-    <h1>Login here</h1>
-  </div>
+  <LoginScreen>
```
paschalidi, line 53: If something is only styles and not a component can we use the Styles suffix please?

src/webapp/src/auth/AuthenticationOptions.tsx
```diff
@@ -1,7 +1,68 @@
+import { Button } from '@small-ads/ui';
 import React from 'react';
+import styled from 'styled-components';
+import { FacebookSquare } from '@styled-icons/boxicons-logos';
```
paschalidi, line 4: Lets use this one please `@styled-icons/boxicons-logos/Facebook`

packages/ui/src/components/Button/index.tsx
```diff
@@ -20,17 +22,21 @@ export function Button({
...
     <ButtonStyles>
       <button
         type={submit ? 'submit' : 'button'}
         disabled={disabled}
         onClick={onClick}
-        className={`${className} ${type} ${size}`}
+        className={`${className} ${type} ${size} ${block}`}
         style={style}
       >
+        <span>{icon}</span>
```
paschalidi, line 39:
```suggestion
        { icon && <span>{icon}</span> }
```

packages/ui/src/components/Button/index.tsx
```diff
@@ -20,17 +22,21 @@ export function Button({
   className = '',
   children,
   onClick,
-  size,
+  size = '',
+  full = false,
+  icon,
 }: ButtonProps) {
+  const block = full ? 'block' : '';
```
paschalidi, line 29: here you want to try this https://styled-components.com/docs/basics#adapting-based-on-props

packages/ui/src/components/Button/button.style.ts
```diff
@@ -59,10 +60,24 @@ export const ButtonStyles = styled.div`
...
   }
+  .large {
+    height: 50px;
+    span {
+      position: absolute;
+      display: flex;
+      -webkit-box-align: center;
+      align-items: center;
+      inset: 0px auto 0px 22px;
+      z-index: 1;
```
paschalidi, line 77: I would rework this. Try to avoid absolute it wont scale well for us. Imagine the case that you have a very long text.

src/webapp/src/auth/AuthenticationOptions.tsx
```diff
@@ -1,7 +1,68 @@
...
+  flex-direction: column;
+`;
+
+const LoginSign = styled.div`
+  font-size: 3rem;
+  letter-spacing: -0.066875rem;
+  font-weight: 700;
+`;
+
+const LoginForm = styled.div`
```
paschalidi, line 27: I dont see any `form` per se. Form is usually html elements. Will end up being a little bit confusing when we actually start having forms.

paschalidi (commented):
Good work, I hit you hard with some of my comments but to be fair most of them are details. You did a good work.

paschalidi:
> I was just trying to solve the four points you gave me. Well, that is not 100% true, I did made a very tiny change in the main.js from Storybook, we were using a deprecated way and I literally just added one @* for our blob to solve it.

Strictly speaking yeah this doesnt belong here. But I am not _that_ strict. 

This is relevant here as you say because this is completely irrelevant. This deprecation I mean but is also one line. 

Now when this rule applies is for large codebase where people start implementing a feature and on the way they find that they want to refactor something and so on.. Then in these case yu have a LOT of noice. But as always is good practise to get used to splitting your work. Colleagues will appreciate you being nitty and taking care of their cognitive load, believe me

src/webapp/src/auth/AuthenticationOptions.tsx
```diff
@@ -1,7 +1,68 @@
...
+    }
+  }
+`;
 
 export const AuthenticationOptions = () => (
-  <div>
-    <h1>Login here</h1>
-  </div>
+  <LoginScreen>
+    <LoginMain>
```
paschalidi, line 54: `LoginMain` is also a bit sketchy

paschalidi:
And dude add descriptions on your PRs. 🚓

packages/ui/src/components/Button/index.tsx
AdrianValdes: Until now I have seen Vercel has 4 types of buttons 'small' | 'medium' | 'large' and another one, the one they use for the `Login` and `SignUp` buttons. That is why I left it blank, so it would be our default button.

 If we don't leave a space we would have to create a new size then, I can call it `regular` and then extract the attributes than now are part of the `button` tag

src/webapp/src/auth/AuthenticationOptions.tsx
AdrianValdes: We need it to increment the [specificity](https://styled-components.com/docs/faqs#how-can-i-override-styles-with-higher-specificity) of the `class-name` inserted by `styled-components`, if we don't do it the styles are going to be overridden by those from the component we are importing from the `ui`

packages/ui/src/components/Button/index.tsx
paschalidi: > That is why I left it blank, so it would be our default button.

Default for me means that if I wanted to use the button component I would not pass a prop regarding the size. 
So would be something like this 

```
<Button fullWidth type="primary"> this is a default button <Button/>
```

How do you think of using default?

AdrianValdes: Exactly, that is our default button:

```
<Button fullWidth type="primary"> this is a default button <Button/>
```

So it doesn't need any size property, the attributes that modify the button are directly in the button tag on our button.styles.ts:
```
button {
    height: 40px;
    ...
    ...
    }
```

src/webapp/src/auth/AuthenticationOptions.tsx
AdrianValdes: You are right, your are right.

src/webapp/src/auth/AuthenticationOptions.tsx
AdrianValdes: I copied the name from Vercel, I understood that `Geist` was like a `ghost` component to make space. Should we name it `GhostSpacer`?

src/webapp/src/auth/AuthenticationOptions.tsx
AdrianValdes: Yes we did, I even did that on purpose. Consistency, consistency, consistency.

src/webapp/src/auth/AuthenticationOptions.tsx
AdrianValdes: It makes the letter closer together so the sign  `Login to Small-ads` will be shorter:

src/webapp/src/auth/AuthenticationOptions.tsx
AdrianValdes: Ok, understood. I will make a note!

src/webapp/src/auth/AuthenticationOptions.tsx
AdrianValdes: Of course! Makes sense

src/webapp/src/auth/AuthenticationOptions.tsx
AdrianValdes: I copied from Vercel :(. I will start renaming their things when they don't make too much sense, thanks!

src/webapp/src/auth/AuthenticationOptions.tsx
AdrianValdes: I used `LoginStyles` trying to follow the previous logic you mentioned. I cannot find a better name that doesn't use *Main*

packages/ui/src/components/Button/index.tsx
AdrianValdes: Right! So we don't have an empty `span` for no reason! I missed it!!! Shit. Thanks, thanks!

packages/ui/src/components/Button/index.tsx
paschalidi: right that makes sense thanks for explaining. Cant we have as default `medium`?

src/webapp/src/auth/AuthenticationOptions.tsx
paschalidi: Cool stuff! Always learning something new with you!

src/webapp/src/auth/AuthenticationOptions.tsx
paschalidi: If Navbar isnt living in the ui then we cant be importing its height from the ui. We can have a `constants.ts` in the Navbar component directory

src/webapp/src/auth/AuthenticationOptions.tsx
paschalidi: `GhostSpacer` sounds cool yeah

packages/ui/src/components/Button/index.tsx
AdrianValdes: Yes, I can refactor what `medium` is now so it looks like our default. Then we need another name for what `medium` is now. It is very similar but the padding is different, I will think and try to find something that makes sense

src/webapp/src/auth/AuthenticationOptions.tsx
paschalidi: Okay fair then. Lets keep it. We only following the design in the end :)

src/webapp/src/auth/AuthenticationOptions.tsx
AdrianValdes: Right, I thought we could export also constants from the `ui`. But I can see how this is better. thanks

src/webapp/src/auth/AuthenticationOptions.tsx
paschalidi: Sorry for being special about those things. But hope you see my logic 👍

packages/ui/src/components/Button/index.tsx
AdrianValdes: How do you find to have `medium` for our default buttons, and `medium-sm` for the buttons that are `medium` now. Reasons?
1. They are bigger than `small` but smaller than `medium`.
2. sm is widely used for buttons sizes (Chakra uses it).
3. The other names I found are not better, `regular`, `medium-small`.

paschalidi: I am confused now to be honest. I think I miss something. I am checking the stories

paschalidi: Okay I think this is my bad here. I didnt understand what you said when you meant there is another button. I thought you were talking about these ones... 
image
Sorry dude. 

okay well now I see what you mean... 

Okay sorry that I missed that bit again. Is a good opportunity to rename them anyway. 

What about `sm` `md` `lg` `xl`?

packages/ui/src/components/Button/index.tsx
```diff
@@ -20,17 +22,21 @@ export function Button({
   className = '',
   children,
   onClick,
-  size,
+  size = 'medium',
+  fullWidth = false,
+  icon,
 }: ButtonProps) {
+  const block = fullWidth ? 'block' : '';
```
paschalidi, line 29: here you want to try this https://styled-components.com/docs/basics#adapting-based-on-props

repeating this one here

src/webapp/src/auth/AuthenticationOptions.tsx
```diff
@@ -1,7 +1,69 @@
...
+  }
+`;
 
 export const AuthenticationOptions = () => (
-  <div>
-    <h1>Login here</h1>
-  </div>
+  <LoginScreenStyles>
+    <LoginStyles>
+      <LoginSign>Login to Small-ads</LoginSign>
```
paschalidi, line 56:
```suggestion
      <LoginSign>Login to upload your ads</LoginSign>
```

packages/ui/src/components/Button/index.tsx
AdrianValdes: Yes, yes. I haven't forgot about that one. I actually read about that one but I avoid it for some other issues that appear,  I am thinking how to solve them this time. But thanks for repeating it. It is better to have it here, easier to see!

packages/ui/src/components/Button/index.tsx
AdrianValdes: I thought something was weird in what we were saying ;). It is all good, of course. We can totally do that and use `sm` `md` `lg` `xl`. The only weird thing is that in that sequence or default button would be `lg` if we consider their height and width. If you are ok with `lg` being our default I don`t mind.

AdrianValdes: If we make `md` our default it would be weird because `md` is going to be bigger than `lg` and that doesn't really make sense I suppose.

packages/ui/src/components/Button/button.style.ts
AdrianValdes: hi dear @paschalidi 
I was trying to find another solution until I checked what they did. 
In this case if we have a very long text the icon would still be separated 22px from the left margin and the auto from (`inset: 0px auto 0px 22px;`) will keep it separated from the text if there is space. Even if the text is super long if the button has space the icon would be in the same place.
I will keep trying to find a better solution. Could you please give me another example in which this way wouldn't work? I am just asking because I think I need it to see what I need to tackle. I just pushed a button with a very long text in case you want to check it out.

packages/ui/src/components/Button/index.tsx
paschalidi: right thanks for explaining ma man. okay lets fo then this `xs` `sm` `md` `lg`. This then would make sense for both of us, or??

packages/ui/src/components/Button/button.style.ts
paschalidi: Well if this works I cant think of anything then :) You good to go. If something pops up we fix it then.

paschalidi:
Ah yeah one more thing. We said that we want for the Navbar to follow the same file structure as the storybook. At the moment this isnt really the case. Can we move the `Navbar.ts` code into the `index.ts` please?
image

packages/ui/src/components/Button/button.style.ts
```diff
@@ -58,11 +55,31 @@ export const ButtonStyles = styled.div`
     line-height: 22px;
     padding: 0px 10px;
```
paschalidi, line 56: I see I have missed these before. Sorry. Lets make these padding. for xm -> 8px for sm -> 12px for md -> 24px and for lg -> 32px

packages/ui/src/components/Button/index.tsx
AdrianValdes: Yes!! That is the best solution, I am the one to have to be sorry for not presenting that solution in my answer, really.

I think I have to be so concentrated new in the things I am learning that I am missing the obvious. But not excuses! More work is what I have to do. Thanks for the patience man!

packages/ui/src/components/Button/button.style.ts
AdrianValdes: It is done, this is also related to the padding being multiples of 4. I will also read about it. Thanks

packages/ui/src/components/Button/index.tsx
AdrianValdes: Hi @paschalidi,
Hi man, this issue is driving me crazy, Every solution I tried throws a new problem. I am assuming we want to use [this](https://styled-components.com/docs/basics#adapting-based-on-props) technique directly in Storybook. I have try different solutions but I cannot solve it. 

What would the right step me to follow now? If I ask for your help should I present what I have tried? It is a lot of code to past here and I know that is not right. Should I push different commits with different failed solutions? Could you point me to the right way of acting in a situation like this one? I know we have had similar situations, but I am trying to get a feeling for what to do in each case.

paschalidi: usually in a working scenario you would pair with someone for 10 ~ 20 minutes. In this situation what I can do is create a https://codesandbox.io/s/pass-in-props-sr258 and I will try to create something similar there.

AdrianValdes: hi @paschalidi ,
Thanks for that man! I wasn't at home and I saw you answered something and I couldn't wait to get here to see what you did. This was the line I couldn't figure out:

`<ButtonStyles isPink={isPink}>`

Now it is so obvious!!!! :D Thanks man

AdrianValdes:
> And dude add descriptions on your PRs. :police_car: 

I will, I will! Thanks for having so much patience with things like that.  It is almost 100% in my head, maybe will happen two more times but then never again!:D

packages/ui/src/components/Button/Button.stories.tsx
```diff
@@ -1,30 +1,52 @@
...
+SmSecondary.args = { type: 'secondary', size: 'sm' };
+
+export const LgSecondary = Template.bind({});
+LgSecondary.args = { type: 'secondary', size: 'lg' };
+
+export const WithIcon = Template.bind({});
+WithIcon.args = {
+  icon: <Facebook size={20} />,
+  style: {
+    width: '220px',
```
paschalidi, line 37: Can we remove this from here and instead have a decorator on the export default? Something like the following
```
decorators: [
    (Story: any) => (
      <Container mt="40px">
        <Story />
      </Container>
    ),
  ],
```

Also can we have stories for each one of the different sizes we are having?

Perhaps we can do something like this for all the stories we are having, what you think?
image

packages/ui/src/components/Button/index.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/jsx-props-no-spreading */
```
paschalidi, line 1: what is this helping us with?

src/webapp/src/auth/AuthenticationOptions.tsx
```diff
@@ -1,7 +1,69 @@
...
-    <h1>Login here</h1>
-  </div>
+  <LoginScreenStyles>
+    <LoginStyles>
+      <LoginSign>Login to upload your ads</LoginSign>
+      <LoginButtons>
+        <GhostSpacer aria-hidden='true' />
+        <FacebookButton
+          size='lg'
+          icon={<Facebook size={20} />}
```
paschalidi, line 61: Can we find a way that the size is standard for our three different buttons. We dont want developers to be passing the `20` around coz in the end some people will be having `21` some `22` some `18` and so on. I think we need to think of 4 defaults and give the user the ability to change them if they needed.

src/webapp/src/components/Navbar/index.ts
```diff
@@ -1 +1,2 @@
 export * from './Navbar';
+export * from './constants';
```
paschalidi, line 2: What about my other comment which is related to this, did you perhaps had the chance to think about it ?

paschalidi (commented):
Looks good to me, few more comments

packages/ui/src/components/Button/Button.stories.tsx
AdrianValdes: image

Yes this is very good, it is a lot better.

AdrianValdes: > Can we remove this from here and instead have a decorator on the export default? Something like the following

Hey man, I am confused about something, what is **this** in that sentence? What I see is line 34 to 37 of the file and if I check Github doesn't say more.

```
WithIcon.args = {
  icon: <Facebook size={20} />,
  style: {
    width: '220px',
```
This is what I see in green.

src/webapp/src/components/Navbar/index.ts
AdrianValdes: Do you mean the one where you mentioned this, right?

> Can we move the Navbar.ts code into the index.ts please? Maybe on a different PR

I created an issue to do it in a different PR, but I can do it here too. I was trying to keep it organized but it is true it is not a bid deal.

src/webapp/src/auth/AuthenticationOptions.tsx
AdrianValdes: It makes sense. So it now the icon have a size by default but the user can also put their own size if they want to.

src/webapp/src/components/Navbar/index.ts
paschalidi: Different Pr is better imo

packages/ui/src/components/Button/Button.stories.tsx
paschalidi: The code changed and changed my selection. One sec.

paschalidi (approved):
Great work dude, thanks for putting this together and for all the effort :) 

2 more things. One is the file structure of the Navbar and the other one is more stories with the icon.

You can tackle the more stories here or on different PR, up to you.

packages/ui/src/components/Button/Button.stories.tsx
AdrianValdes: I already did it, but I will merge and open a new PR. I want to follow was is best for you/us!

PR #58, opened 2021-03-07
refactor: improve icon positioning

AdrianValdes:
hi @paschalidi,

1. When I was showing different icons sizes in each story I realized this is a better way of handling our icon. So now instead of having the icon always completely to the left, we will have it next to the button-text (how Chakra does it). Then we will use `position: absolute` for the specific cases if we want to do something similar to what Vercel does for the login buttons.
   
  You were right all along with changing that! What a surprise!!! :D Who am I to doubt you?! :D

2. I add all the sizes of a button for each story. I chose to display the button one top of each other, instead of one next to each other, but we can change it of course. Would you like me to add some more story?

### DIFF
--- a/packages/ui/src/components/Button/Button.stories.tsx
+++ b/packages/ui/src/components/Button/Button.stories.tsx
@@ -8,7 +8,7 @@ type Sizes = ['xs', 'sm', 'md', 'lg'];
 const sizes: Sizes = ['xs', 'sm', 'md', 'lg'];
 
 const ButtonsSeparator = styled.div`
-  margin-bottom: 15px;
+  margin-bottom: 16px;
 `;
 
 const Template: Story<ButtonProps> = (args) => {
--- a/packages/ui/src/components/Button/Button.stories.tsx
+++ b/packages/ui/src/components/Button/Button.stories.tsx
@@ -2,43 +2,40 @@ import React from 'react';
 import { Meta, Story } from '@storybook/react';
 import { Button, ButtonProps } from '.';
 import { Facebook } from '@styled-icons/boxicons-logos/Facebook';
+import styled from 'styled-components';
 
-const Template: Story<ButtonProps> = (args) => (
-  <Button {...args}>Call to action </Button>
-);
+type Sizes = ['xs', 'sm', 'md', 'lg'];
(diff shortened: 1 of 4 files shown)


AdrianValdes:
On another note:

This code bothers me a little here

```
type Sizes = ['xs', 'sm', 'md', 'lg'];
const sizes: Sizes = ['xs', 'sm', 'md', 'lg'];

const ButtonsSeparator = styled.div`
  margin-bottom: 15px;
`;
```

But I didn't want to create another file for that because that would be to break our structure. Also it is not that much and I saw the people from Storybook doing it. I can always change it if you think is best, I just don't want you to think I didn't care :D

packages/ui/src/components/Button/Button.stories.tsx
```diff
@@ -2,55 +2,51 @@ import React from 'react';
...
-const Template: Story<ButtonProps> = (args) => (
-  <Button {...args}>Call to action </Button>
-);
+import styled from 'styled-components';
+
+type Sizes = ['xs', 'sm', 'md', 'lg'];
+const sizes: Sizes = ['xs', 'sm', 'md', 'lg'];
+
+const ButtonsSeparator = styled.div`
+  margin-bottom: 15px;
```
paschalidi, line 11: I would encourage you to start using derivatives of 4px. ;)

paschalidi (commented):
image
Looks good overall thanks for the change! Look at the image, looks like something is odd

packages/ui/src/components/Button/Button.stories.tsx
AdrianValdes: Shit for the margins too!! My baad, So padding, margins, what else? :D

AdrianValdes:
That is weird, I don't see that problem with the padding or the icon.

packages/ui/src/components/Button/Button.stories.tsx
paschalidi: Not so important ;)

paschalidi:
Looks good now, was late last night perhaps I changed the size without me realising it..

paschalidi (approved):
Dont go crazy I think I must have changed the size of the icon somehow without me realising it! 

Good work again!

PR #61, opened 2021-03-08
refactor: :construction: improve useSelector usage

AdrianValdes:
Hi @paschalidi,

Now that we want to start introducing `redux` functionality in the login page I want to propose a small change in the way we consume the data from our store. 

Reasons:
1. It makes our code cleaner.
2. It is the way the `redux-toolkit` [recommend ](https://redux-toolkit.js.org/tutorials/typescript#application-usage)to do it, even when using Typescript.

Description: 
By adding one line of code `export const selectUser = (state: RootState) => state.user` we are able of using the data from the store easier:  `const { authStatus } = useSelector(selectUser)`.

Disadvantages:
The linter complains so we would have to disable the `import/no-cycle` rule for the line, the file or in general, whatever you think it is best. Apparently it is not a big deal to disable it for the file in [our](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-cycle.md) case.

Like always if you think this is not a good idea, it is no a problem at all. Just let me know. It is very easy to reverse it at this point.


paschalidi:
Made a suggestion here https://github.com/small-ads/core/pull/62, lets chat about the pros and the cons

AdrianValdes:
I saw the comment, you are totally right. I should have found the work around myself. What you suggested is much more cleaner, I agree is better to **not** disable the rules. I have to get better at thinking about finding a work around when I need it :D, Thanks man!

src/webapp/src/auth/Auth.tsx
```diff
@@ -2,8 +2,8 @@ import React from 'react';
 import styled from 'styled-components';
 import { useDispatch, useSelector } from 'react-redux';
 import { AuthStatuses } from './types';
-import { RootState } from '../reducers';
 import { signInUser, signOutUser } from './asyncAuthActions';
+import { selectUser } from "./selectors"
```
paschalidi, line 6:
```suggestion
import { selectUser } from './selectors'
```
double quotes arent what we use usually, what makes us want to have them here?

AdrianValdes: I have no idea why they are appearing there! It is driving me crazy, I have singleQuotes: true in every possible field.

AdrianValdes:
I even have single quotes in my file now :thinking:

paschalidi (approved):
LGTM!

Issue #40, opened 2021-03-02
user can login


paschalidi:
Acceptance criteria:
- [x]  There is a menu on the top of the screen where you can click Login. 
- [x] When you click login you are navigated to another route. There you see all the login options we provide - facebook, google, maybe more. Example of flow https://vercel.com/
- [x] When the user is logged in then they can click a button and logout. Session is cleared and the user could login again.
- [x] When user logs in they see the home page route where there is default list of ads

AdrianValdes:
hi @paschalidi ,
Do you think it makes sense to build the `navbar` as a component in Storybook in a project like this? I think it may no be necessary but I am not sure.

paschalidi:
You can decide. That I think as you say is on the edge. It doesnt have to be though in the components since it is not a reusable component. 

Very much depends on the agreement we wanna make on this. So as I said, you can choose what is better

PR #67, opened 2021-03-13
feat: user can login from the navbar

AdrianValdes:
User can login from the Navbar and go to the `/dashboard` route (this is the name Vercel uses, but I think you are not convinced, let my know if you have a better idea).

### DIFF
--- a/src/webapp/src/App.tsx
+++ b/src/webapp/src/App.tsx
@@ -13,7 +13,7 @@ export const App = () => (
       <Navbar />
       <Switch>
         <Route path='/login' component={AuthenticationOptions} />
-          <Route path='/feed' component={Feed} />
+        <Route path='/feed' component={Feed} />
         <Route exact path='/' component={Auth} />
       </Switch>
     </Router>
--- a/src/webapp/src/App.tsx
+++ b/src/webapp/src/App.tsx
@@ -4,6 +4,7 @@ import { BrowserRouter as Router, Switch, Route } from 'react-router-dom';
 import { Navbar } from './components/Navbar';
 import { AuthenticationOptions } from './auth/AuthenticationOptions';
 import { Auth } from './auth/Auth';
+import { Dashboard } from './home/Dashboard';
 
 export const App = () => (
   <>
@@ -12,6 +13,7 @@ export const App = () => (
       <Navbar />
       <Switch>
         <Route path='/login' component={AuthenticationOptions} />
+        <Route path='/dashboard' component={Dashboard} />
         <Route exact path='/' component={Auth} />
       </Switch>
     </Router>
--- a/src/webapp/src/App.tsx
+++ b/src/webapp/src/App.tsx
@@ -4,7 +4,7 @@ import { BrowserRouter as Router, Switch, Route } from 'react-router-dom';
 import { Navbar } from './components/Navbar';
 import { AuthenticationOptions } from './auth/AuthenticationOptions';
 import { Auth } from './auth/Auth';
-import { Dashboard } from './home/Dashboard';
+import { Feed } from './home/Feed';
 
 export const App = () => (
   <>
@@ -13,7 +13,7 @@ export const App = () => (
       <Navbar />
       <Switch>
         <Route path='/login' component={AuthenticationOptions} />
-        <Route path='/dashboard' component={Dashboard} />
+          <Route path='/feed' component={Feed} />
         <Route exact path='/' component={Auth} />
       </Switch>
     </Router>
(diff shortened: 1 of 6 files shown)


src/webapp/src/auth/AuthenticationOptions.tsx
```diff
@@ -69,20 +74,32 @@ const GoogleButton = styled(Button)`
...
-      </LoginButtons>
-    </LoginStyles>
-  </LoginScreenStyles>
-);
+export const AuthenticationOptions = () => {
+  const dispatch = useDispatch();
+  const { authStatus } = useSelector(selectUser);
+
+  if (authStatus === AuthStatuses.loggedIn) {
+    return <Redirect to='/dashboard' />;
```
paschalidi, line 82: Nice thinking.

src/webapp/src/components/Navbar/index.tsx
```diff
@@ -1,23 +1,39 @@
...
-        >
-          Login
-        </Button>
+        {!authStatus && (
+          <Button
+            disabled={Boolean(route?.isExact)}
+            size='sm'
+            onClick={() => {
+              history.push('/login');
+            }}
```
paschalidi, line 27: We have two onClick handlers in this file. One has an inline function and one is not. Perhaps we can decide on a convention and follow it accross the project. What you think?

src/webapp/src/components/Navbar/index.tsx
```diff
@@ -1,23 +1,39 @@
...
+          >
+            Login
+          </Button>
+        )}
+        {Boolean(authStatus) && (
```
paschalidi, line 32: A bit confused here. `authStatus` can be 
```
{
  notLoggedIn,
  loggedIn,
  loading,
}
``` 
How is that in the case `notLoggedIn` we want to show the button? I undestant that this might be impossible now because you are redirecting but still this is a bit cryptic for me here. Can we be explicit in this statement?

src/webapp/src/components/Navbar/index.tsx
```diff
@@ -1,23 +1,39 @@
...
-        <Button
-          disabled={Boolean(route?.isExact)}
-          size='sm'
-          onClick={() => {
-            history.push('/login');
-          }}
-        >
-          Login
-        </Button>
+        {!authStatus && (
```
paschalidi, line 21: I would say the same is here.

src/webapp/src/App.tsx
```diff
@@ -12,6 +13,7 @@ export const App = () => (
       <Navbar />
       <Switch>
         <Route path='/login' component={AuthenticationOptions} />
+        <Route path='/dashboard' component={Dashboard} />
```
paschalidi, line 16: I would say this page for me a feed of ads. What you think of renaming this to a little?

paschalidi (commented):
Hey good work thanks for the nice small PR! 
Code looks great as well. 

Few points I could think of only. 

1. What I refresh the page the login status is being lost. What is your plan in making this persistent?
2. I want to start implementing a `Grid`. When you think this will fit in?

src/webapp/src/components/Navbar/index.tsx
AdrianValdes: Of course, it makes only sense to be consistent. I am going to go with **not** using inline functions because if the functions are too long it can obscure the code.

src/webapp/src/App.tsx
AdrianValdes: Perfect, I was missing a name like that. Do you prefer `AdsFeed` or something longer like `FeedOfAds`?

paschalidi: perhaps just `feed` might be good

AdrianValdes: > What I refresh the page the login status is being lost. What is your plan in making this persistent?

I implemented something using firebase `auth.onAuthStateChanged`, didn't push it now to keep the commit clean. As soon as we polish what you comment I will show you what I thought of.

AdrianValdes:
> What I refresh the page the login status is being lost. What is your plan in making this persistent?

I implemented something using firebase `auth.onAuthStateChanged`, didn't push it now to keep the commit clean. As soon as we polish what you comment I will show you what I thought of.

AdrianValdes:
> I want to start implementing a Grid. When you think this will fit in?

hehe. That is going to be tough and fun for me. Maybe after we finish with issue #40?

src/webapp/src/App.tsx
```diff
@@ -12,6 +13,7 @@ export const App = () => (
       <Navbar />
       <Switch>
         <Route path='/login' component={AuthenticationOptions} />
+          <Route path='/feed' component={Feed} />
```
paschalidi, line 16:
```suggestion
        <Route path='/feed' component={Feed} />
```

paschalidi (approved):
Looks good. Good work as always dude. Can you fix that bit over there?

PR #68, opened 2021-03-14
fix: eslint improvement

paschalidi:
Adds `'plugin:prettier/recommended'` in the list. This will make eslint recognise errors when our jsx is not formatted right as it should have done here for example https://github.com/small-ads/core/pull/67/commits/5f8660c4807945842b0c4dc03e15b6481cd5ca99#diff-746224c245c31b8e0092392bf7c22c55cc0779a84be1f9ab6715d35c3f77a7ebR16

### DIFF
--- a/packages/eslint-config-small-ads/index.js
+++ b/packages/eslint-config-small-ads/index.js
@@ -7,6 +7,7 @@ module.exports = {
     'plugin:import/recommended',
     'plugin:import/typescript',
     'plugin:react-hooks/recommended',
+    'plugin:prettier/recommended',
     'airbnb-typescript',
     'prettier',
     'prettier/react',
--- a/packages/ui/src/setupTests.ts
+++ b/packages/ui/src/setupTests.ts
@@ -2,4 +2,4 @@
 // allows you to do things like:
 // expect(element).toHaveTextContent(/react/i)
 // learn more: https://github.com/testing-library/jest-dom
-import "@testing-library/jest-dom";
+import '@testing-library/jest-dom';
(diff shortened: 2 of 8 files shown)


AdrianValdes:
Oh, that happened because I just added a line using GitKraken, so when I add a hunk that could change the format, I have to keep that in mind. Sorry.

AdrianValdes:
Why is the linter complaining about  `error  Replace `"@testing-library/jest-dom"` with `'@testing-library/jest-dom'`  prettier/prettier`
I don't get it, we have single quotes there.

paschalidi:
@AdrianValdes check this out

AdrianValdes:
So we are going to have double quotes for the components properties `<button type="button" onClick={handleSignIn}>` now? I am just asking because that was the behavior I had locally and I had to add a .prettierrc to modify it so I want to know if I have to delete it.

paschalidi:
I think this is the intended behaviour of the linter however because we werent having this plugin on the linter wasnt catching it.

What are the issues you are seeing?

AdrianValdes:
> What are the issues you are seeing?
No problem, all good, I wanted to know if we wanted this behavior. I had that behavior before, maybe you remember I was making some PRs with the double-quotes there. I had to fight against it creating a separated `.pretterrc`, but no problem, I will delete it and I think I will have this behavior back.

paschalidi:
Can you make a review when you have the time

AdrianValdes (approved):
Thanks for the change. It is going to keep the code clean

PR #69, opened 2021-03-14
feat: logged in user is persistent

AdrianValdes:
hi @paschalidi,
Do you think this is the good way of making the user persistent in our app? I read a bit about it and a few people were doing it like this. I don't think is bad, but I am not sure if it is the best we can do now.
Also, I realize we need some real loading component too, maybe after making the Grid?


paschalidi:
Can we try a spinner of those for this PR at least? Looks like half finished now to me and is difficult to imagine how it would be working in the end

AdrianValdes:
I am working on it, today I had classes an tutoring so I had little time :(.

Now I am wondering how to handle the rendering of the component, other sites seem to delay the rendering until they check if the user was already logged-in, but I am not sure how to do it. I will keep searching to try to show you some idea tomorrow.

paschalidi (approved):
Hey sorry for my very late reply here. 

I forgot to review this one on time and then I forgot about it. 

As usual good work on this!

